PR #13, opened 2023-03-09
Add IPAM support

rabi:
Initial ipam support to use in dataplane-operator.


stuggi:
adding @olliewalsh who also wanted to work on the ipam topic

apis/network/v1beta1/openstacknet_types.go
```diff
@@ -0,0 +1,153 @@
...
+import (
+	"github.com/openstack-k8s-operators/lib-common/modules/common/condition"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// OpenStackNetSpec defines the desired state of OpenStackNet
+type OpenStackNetSpec struct {
+	// +kubebuilder:validation:Optional
+	//  List of subnets
+	Subnets []OpenStackSubnet `json:"subnets,omitempty"`
```
stuggi, line 28: wondering if we need this distinction openstacknet -> opernstacksubnet list where a net has multiple subnets? or wouldn't it be enough to just have openstacknet?

apis/network/v1beta1/openstackipset_types.go
```diff
@@ -0,0 +1,101 @@
...
+	// Fixed Ip
+	FixedIP string `json:"fixedIP,omitempty"`
+}
+
+// OpenStackIPSetSpec defines the desired state of OpenStackIPSet
+type OpenStackIPSetSpec struct {
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=false
+	// VIP flag to indicate ipset is a request for a VIP
+	VIP bool `json:"vip"`
```
stuggi, line 44: not sure if we need this for podified ctlplane. in OSPDO it was required to generate the needed env files. now I think we could just request an IP, or did I miss a use case?

apis/network/v1beta1/openstacknet_types.go
```diff
@@ -0,0 +1,153 @@
...
+	Nexthop string `json:"nexthop"`
+}
+
+// IPReservation contains an IP and Deleted flag
+type IPReservation struct {
+	// +kubebuilder:validation:Required
+	IP string `json:"ip"`
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=false
+	VIP bool `json:"vip"`
```
stuggi, line 101: as mentioned above, we probably don't need to indicated if the reservation is for a VIP?

apis/network/v1beta1/openstackipset_types.go
```diff
@@ -0,0 +1,101 @@
...
+
+	// Conditions - conditions to display in the OpenShift GUI, which reflect CurrentState
+	// +kubebuilder:validation:Optional
+	Conditions condition.Conditions `json:"conditions,omitempty"`
+	// Important: Run "make" to regenerate code after modifying this file
+}
+
+//+kubebuilder:object:root=true
+//+kubebuilder:subresource:status
+
```
stuggi, line 69: might want to add some columns to display on the cli, min status and message

```
//+kubebuilder:printcolumn:name="Status",type="string",JSONPath=".status.conditions[0].status",description="Status"
//+kubebuilder:printcolumn:name="Message",type="string",JSONPath=".status.conditions[0].message",description="Message"
```

controllers/network/openstackipset_controller.go
```diff
@@ -0,0 +1,264 @@
...
+	// Build AssignIPDetails
+	networks := instance.Spec.Networks
+
+	if instance.Status.IPAddresses == nil {
+		instance.Status.IPAddresses = make(map[string]networkv1beta1.IPReservation)
+	}
+
+	//find the the openstacknet objects
+	for _, network := range networks {
+		if _, ok := instance.Status.IPAddresses[network.Name]; ok {
```
stuggi, line 128: should we check if the reservation is still on the osnet, or is it save to say we stored it once we expect it to be there? if we change the reservation to be a map[string]IPReservation, this would be easy to check

apis/network/v1beta1/openstacknet_types.go
```diff
@@ -0,0 +1,153 @@
...
+	// Gateway optional gateway for the network
+	Gateway string `json:"gateway,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// Routes, list of networks that should be routed via network gateway.
+	Routes []Route `json:"routes,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// IP Reserverations for Subnet
+	Reservations []IPReservation `json:"reservations,omitempty"`
```
stuggi, line 70: could also be a map[string]IPReservation where the key is the hostname of the consumer. would make it easy to validate and also for delete a reservation

controllers/network/openstackipset_controller.go
```diff
@@ -0,0 +1,264 @@
...
+				ipReservation.Gateway = subnet.Gateway
+				ipReservation.Routes = subnet.Routes
+
+				aid.Reservelist = updatedReservations
+				osnet.Spec.Subnets[i].Reservations = append(osnet.Spec.Subnets[i].Reservations, ipReservation)
+
+				err = r.Update(ctx, osnet)
+				if err != nil {
+					return ctrl.Result{}, err
+				}
```
stuggi, line 167: I think we want to use Patch() instead of update that on reconciliation the osnet only gets updated if this controller did a change to the current object version.

controllers/network/openstackipset_controller.go
```diff
@@ -0,0 +1,264 @@
...
+			return ctrl.Result{}, err
+		}
+
+		// Remove the finalizers from networks
+		if controllerutil.RemoveFinalizer(osnet, fmt.Sprintf("%s-%s", helper.GetFinalizer(), instance.Name)) {
+			err := r.Update(ctx, osnet)
+			if err != nil {
+				return ctrl.Result{}, err
+			}
+		}
```
stuggi, line 250: should change this to only have a single patch call if possible

apis/network/v1beta1/openstacknet_types.go
```diff
@@ -0,0 +1,153 @@
...
+	// +kubebuilder:validation:Optional
+	Subnets []OpenStackSubnet `json:"subnets,omitempty"`
+
+	// Status Conditions
+	// +kubebuilder:validation:Optional
+	Conditions condition.Conditions `json:"conditions,omitempty"`
+}
+
+//+kubebuilder:object:root=true
+//+kubebuilder:subresource:status
```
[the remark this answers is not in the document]
rabi: ack

config/manager/kustomization.yaml
```diff
@@ -4,5 +4,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: controller
+  newName: quay.io/rabi/infra-operator
```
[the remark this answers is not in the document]
rabi: yes, it just updates that. we need to ignore the file like https://github.com/openstack-k8s-operators/glance-operator/blob/master/.gitignore#L23

apis/network/v1beta1/openstacknet_types.go
rabi: You mean assume one net one subnet?  What about https://github.com/openstack/tripleo-heat-templates/blob/master/network-data-samples/routed-networks.yaml and dual stack?

apis/network/v1beta1/openstackipset_types.go
rabi: I added that to distinguish, if there is a OpenStackIPSet request for a set of vips i.e not associated with a node. If there is no requirement I can remove it later.

apis/network/v1beta1/openstackipset_types.go
rabi: ack

controllers/network/openstackipset_controller.go
rabi: We can surely add more checks for cases like that, but I did not want to make the CRDs overly complex. Like OpenStackNet to know who has requested/reserved it.

apis/network/v1beta1/openstacknet_types.go
rabi: We can, but as commented earlier I wanted not to make it too complex with details about who reserved an IP.

controllers/network/openstackipset_controller.go
rabi: I'll check if patch works and update accordingly.

controllers/network/openstackipset_controller.go
rabi: commented earlier.

controllers/network/openstackipset_controller.go
```diff
@@ -0,0 +1,264 @@
...
+
+		if controllerutil.AddFinalizer(osnet, fmt.Sprintf("%s-%s", helper.GetFinalizer(), instance.Name)) {
+			err = r.Update(ctx, osnet)
+			if err != nil {
+				r.Log.Info("Failed Adding finalizer")
+				return ctrl.Result{}, err
+			}
+		}
```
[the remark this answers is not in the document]
stuggi: ah right there was something. I can remember now, so this might be needed

apis/network/v1beta1/openstackipset_types.go
stuggi: sounds good

controllers/network/openstackipset_controller.go
stuggi: we did it like that in ospdo and just marked remove nodes/entries as deleted so that you could replace a node and get the same reservation. we had a dedicated parameter to always clean reservations on delete. but we could also say if you want a behavior like that, always request a specific IP in the request. either way works

controllers/network/openstackipset_controller.go
stuggi: if it does not work, we should use deepequal to check if there was a change and only call the update if there is a change

rabi: I changed to use patch, but did not notice any benefits and we've already got checks to not update when already reserved.

stuggi:
@rabi can you please put this into draft until we decided which way to go

PR #14, opened 2023-03-09
Change name of memcached configmap

bshephar:
When we create a Memcached CR instance for individual services, then we should ideally append -memcached to the name of the configmap to avoid conflicing with the service configmaps. This change appends the -memcached to the configmap name accordingly.

### DIFF
--- a/controllers/memcached/memcached_controller.go
+++ b/controllers/memcached/memcached_controller.go
@@ -194,7 +194,7 @@ func (r *Reconciler) generateConfigMaps(
 	cms := []util.Template{
 		// ConfigMap
 		{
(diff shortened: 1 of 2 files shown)


bshephar:
This come up as a potential issue while implementing memcached for the Horizon operator. With the objective being to create an instance of the Memcached CR called "horizon". Consequently, the resulting configmap would be called `horizon-config-data` and conflict with the configmap used by Horizon. 

This change ensures that when any service operator attempts to create an instance of memcached, the resulting memcached related configmap will be identifiable by name and not conflict with existing configmaps in use.

gibizer (commented):
looks good to me.

dciabrin:
/lgtm
Thanks

PR #22, opened 2023-04-05
Webhooks for default images

abays:
Default container images are now handled via env vars and internal webhooks. This defaulting logic will in turn later be integrated with OpenStack Operator via webhooks there that use the new defaulting mechanism (while at the same time disabling the Infra service operator internal webhooks in that context).


apis/client/v1beta1/openstackclient_webhook.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+// Default - set defaults for this OpenStackClient spec
+func (spec *OpenStackClientSpec) Default() {
+	if spec.ContainerImage == "" {
+		spec.ContainerImage = openStackClientDefaults.ContainerImageURL
+	}
+}
+
+// TODO(user): change verbs to "verbs=create;update;delete" if you want to enable deletion validation.
```
dciabrin, line 73: Is this comment still required?

config/manager/manager.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 kind: Namespace
 metadata:
   labels:
-    control-plane: controller-manager
+    control-plane: infra-controller-manager
```
dciabrin, line 5: I'm not sure what this renaming does, and whether it has to match what is defined in config/default/manager_webhook_patch.yaml?
```
metadata:
  name: controller-manager
  namespace: system
```

Makefile
```diff
@@ -309,3 +311,20 @@ gowork: ## Generate go.work file to support our multi module repository
...
+# $oc delete -n openstack mutatingwebhookconfiguration/mopenstackclient.kb.io
+# $oc delete -n openstack validatingwebhookconfiguration/vmemcached.kb.io
+# $oc delete -n openstack mutatingwebhookconfiguration/mmemcached.kb.io
+SKIP_CERT ?=false
+.PHONY: run-with-webhook
+run-with-webhook: export INFRA_CLIENT_IMAGE_URL_DEFAULT=quay.io/tripleozedcentos9/openstack-tripleoclient:current-tripleo
+run-with-webhook: export INFRA_MEMCACHED_IMAGE_URL_DEFAULT=quay.io/tripleozedcentos9/openstack-memcached:current-tripleo
+run-with-webhook: manifests generate fmt vet ## Run a controller from your host.
+	/bin/bash hack/configure_local_webhook.sh
+	go run ./main.go
```
dciabrin, line 330: could you replicate line 123 to keep the METRICS_PORT and HEALTH_PORT overrides?

dciabrin (changes requested):
This looks good to me. I only have minor nits.
I'm also wondering if we want to keep the TODO comments on the generated files for guidance, or if they are just unnecessary at this point?

apis/client/v1beta1/openstackclient_webhook.go
abays: Across the operators I have been leaving these `TODO` comments from the webhook scaffolding in-place because they pertain to the validation webhooks, which are now present but haven't been utilized yet (this PR only adds actual logic for  the mutating/defaulting webhook, but see here [1] why both mutating and validating scaffolding are added at the same time, even if we're only using one type for the moment).  I can remove the comment, but it might be useful information for anyone who is looking to actually add validation logic.

[1] https://github.com/openstack-k8s-operators/docs/blob/main/webhooks.md#adding-webhooks-to-an-operator

config/manager/manager.yaml
abays: This change is for the label, which does not pertain to the `metadata.name`.  We are repeating this pattern of altering the label `control-plane: controller-manager` to something more specific like `control-plane: infra-controller-manager` because this label is used by the webhook service to direct CR API traffic to the proper webhook endpoint.  Without this change, all operator webhook servers have an identical label, and k8s gets confused and routes traffic for CRs to the wrong associated `controller-manager` webhook server.

Makefile
abays: Ack, done

apis/client/v1beta1/openstackclient_webhook.go
dciabrin: Clear, thanks for the pointer

config/manager/manager.yaml
dciabrin: Oh alright. Thanks for the explanation.

dciabrin (approved):

PR #24, opened 2023-04-14
Use golang 1.19 in CI

kajinamit:
Now this operator is developed with golang 1.19 instead of 1.18, so we should use the correct version.


stuggi (approved):
/lgtm

PR #25, opened 2023-04-18
openstackclient: Get rid of kolla_start

kajinamit:
... so that we don't need to allow the cloud-admin users to use sudo. Usage of kolla_start is redundant because we have no config files copied into the container by the command.


pkg/openstackclient/funcs.go
```diff
@@ -54,8 +54,10 @@ func ClientPod(
 	clientPod.Spec.ServiceAccountName = ServiceAccount
 	clientPod.Spec.Containers = []corev1.Container{
 		{
-			Name:  "openstackclient",
-			Image: instance.Spec.ContainerImage,
+			Name:    "openstackclient",
+			Image:   instance.Spec.ContainerImage,
+			Command: []string{"/bin/sleep"},
+			Args:    []string{"infinity"},
 			SecurityContext: &corev1.SecurityContext{
```
kajinamit, line 61: We can also remove these and anyuid, but I'd leave it for follow-up.

gibizer:
Looks OK to me but we have no e2e deploy job for this repo yet. Have you deployed this change?

gibizer (approved):
I successfully run this alone as well as together with the tcib change in https://github.com/openstack-k8s-operators/openstack-operator/pull/232

stuggi (approved):
/lgtm

PR #31, opened 2023-04-24
[DNS] Add DNSMasq and DNSData CRD to manage DNS entries

stuggi:
* DNSMasq CRD allows creating a dnsmasq DNS serverinstance and expose
    the service via ExternalEndpoints and metallb to the edpm nodes on
    isolated networks.
    
    Using the DNSMasqOptions the instance can be customized from the default
    config started.
    
    Hosts information is being added via configmaps which have a label with
    {dnsmasqhosts: dnsdata (can be customized via DNSDataLabelSelectorValue)}
    set.
    
* DNSData CRD allows to register hosts to be resolved in the dnsmasq
    instance. The CRD creates the required configmap with the label as
    described above.
    
* Additionally there is a service_controller.go with no CRD. It watches
    for LoadBalancer services in the same namespace of a DNSMasq instance and
    creates a DNSData CR to registers service endpoints which have an annotation
    set with {"dnsmasq.network.openstack.org/hostname": <hostname>}.
    
Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/241
Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/258
Jira: [OSP-23053](https://issues.redhat.com//browse/OSP-23053)


stuggi:
need to run a full e2e test and install_yamls integration

config/samples/network_v1beta1_dnsmasq.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: network.openstack.org/v1beta1
+kind: DNSMasq
+metadata:
+  name: dnsmasq
+spec:
+  containerImage: quay.io/tripleozedcentos9/openstack-neutron-server:current-tripleo
```
stuggi, line 6: TODO: own image

apis/memcached/v1beta1/memcached_types.go
```diff
@@ -73,12 +73,15 @@ func (instance Memcached) IsReady() bool {
...
 // RbacNamespace - return the namespace
-func (instance Memcached) RbacNamespace() string {    return instance.Namespace
+func (instance Memcached) RbacNamespace() string {
+	return instance.Namespace
 }
 
 // RbacResourceName - return the name to be used for rbac objects (serviceaccount, role, rolebinding)
 func (instance Memcached) RbacResourceName() string {
-    return "memcached-" + instance.Name
+	return "memcached-" + instance.Name
```
stuggi, line 86: only to make linter happy. can extract this to its own PR

pkg/dnsmasq/deployment.go
```diff
@@ -0,0 +1,192 @@
...
+		dnsmasqCmd = append(dnsmasqCmd, "--log-debug")
+		dnsmasqCmd = append(dnsmasqCmd, "--bind-interfaces")
+		dnsmasqCmd = append(dnsmasqCmd, "--listen-address=$(POD_IP)")
+		dnsmasqCmd = append(dnsmasqCmd, "--port "+strconv.Itoa(int(DNSPort)))
+		// log to stdout
+		dnsmasqCmd = append(dnsmasqCmd, "--log-facility=-")
+		// dns
+		dnsmasqCmd = append(dnsmasqCmd, "--no-hosts")
+		dnsmasqCmd = append(dnsmasqCmd, "--domain-needed")
+		dnsmasqCmd = append(dnsmasqCmd, "--no-resolv")
```
stuggi, line 100: right now you have to configure an upstream dns server via spe.options in the dnsmasq CR. would it be ok to the ocp cluster one set in the resolv.conf?

controllers/network/service_controller.go
```diff
@@ -0,0 +1,198 @@
...
+
+// getServiceDNSData -
+func (r *ServiceReconciler) getServiceDNSData(
+	ctx context.Context,
+	namespace string,
+) (map[string]networkv1.DNSHost, error) {
+	svcDNSHosts := map[string]networkv1.DNSHost{}
+
+	// get all services from the namespace triggered the reconcile
+	svcList := &corev1.ServiceList{}
```
stuggi, line 116: could do similar with openstackdataplanenodes if there are CRs for each node.

apis/network/v1beta1/dnsdata_types.go
```diff
@@ -0,0 +1,88 @@
...
+
+	// +kubebuilder:validation:Required
+	// Hostnames for the IP address.
+	Hostnames []string `json:"hostnames"`
+}
+
+// DNSDataSpec defines the desired state of DNSData
+type DNSDataSpec struct {
+	// +kubebuilder:validation:Optional
+	Hosts []DNSHost `json:"hosts,omitempty"`
```
abays, line 40: It's weird that having both `Optional` and `omitempty` doesn't fail `operator-lint` in `pre-commit` in Prow.  I thought we checked for that.  Maybe this repo is using an older `operator-lint` version.

apis/network/v1beta1/dnsmasq_types.go
```diff
@@ -0,0 +1,174 @@
...
+type DNSMasqOption struct {
+	// +kubebuilder:validation:Enum=server;rev-server;srv-host;txt-record;ptr-record;rebind-domain-ok;naptr-record;cname;host-record;caa-record;dns-rr;auth-zone;synth-domain
+	Key    string   `json:"key"`
+	Values []string `json:"values"`
+}
+
+// DNSMasqSpec defines the desired state of DNSMasq
+type DNSMasqSpec struct {
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default="quay.io/podified-antelope-centos9/openstack-neutron-server:current-podified"
```
abays, line 44: This should eventually be removed in favor of the env var defaults implemented by webhooks

stuggi: I have added web hook for this. so I think I can remove it

apis/network/v1beta1/dnsdata_types.go
stuggi: isn't that only if there is a default value specified? https://github.com/gibizer/operator-lint/tree/main/linters/crd/C003

apis/network/v1beta1/dnsmasq_types.go
stuggi: done

apis/network/v1beta1/dnsdata_types.go
abays: Ah, I did not know that.  Fair enough.

apis/network/v1beta1/dnsmasq_types.go
```diff
@@ -0,0 +1,172 @@
...
+	// NodeSelector here acts as a default value and can be overridden by service
+	// specific NodeSelector Settings.
+	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// ExternalEndpoints, expose a VIP using a pre-created IPAddressPool
+	ExternalEndpoints []MetalLBConfig `json:"externalEndpoints,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default="dnsdata"
```
abays, line 72: Should this default be the same as the value set on L31?

pkg/dnsmasq/const.go
```diff
@@ -0,0 +1,29 @@
...
+limitations under the License.
+*/
+
+package dnsmasq
+
+const (
+	// ServiceName -
+	ServiceName = "dnsmasq"
+	// ServiceAccount -
+	ServiceAccount = "infra-operator-dnsmasq"
```
dprince, line 22: How is this getting used. And are we generating an account for this in each namespace then?

stuggi: good catch, thanks. missed this part to integrate with the new rbac way. will fix it

apis/network/v1beta1/dnsmasq_types.go
stuggi: no, L31 is the key of the label. this is the value. its probably confusing that in L28 I call it AnnotationHostnameKey, but in L31 don't name it key. I'll change that to be consistent

controllers/network/dnsmasq_controller.go
```diff
@@ -0,0 +1,500 @@
...
+	p := predicate.Funcs{
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			// The object doesn't contain label networkv1.DNSDataLabelSelectorKey, so the event will be
+			// ignored.
+			if _, ok := e.ObjectOld.GetLabels()[networkv1.DNSDataLabelSelectorKey]; !ok {
+				return false
+			}
+
+			return e.ObjectOld != e.ObjectNew
+		},
```
abays, line 239: So is this an improvement on the old pattern where we just check for the labels in the watch function itself?

controllers/network/dnsdata_controller.go
```diff
@@ -0,0 +1,228 @@
...
+		r.Log.Info(fmt.Sprintf("Input maps hash %s - %s", common.InputHashName, hash))
+	}
+
+	instance.Status.Conditions.MarkTrue(condition.ServiceConfigReadyCondition, condition.InputReadyMessage)
+
+	r.Log.Info("Reconciled Service successfully")
+	return ctrl.Result{}, nil
+}
+
+// generateServiceConfigMaps - create create configmap hosts file
```
abays, line 194: nit: `s/create create/create/`

controllers/network/dnsmasq_controller.go
stuggi: right, this is to limit not required reconciliation.

apis/network/v1beta1/dnsmasq_types.go
stuggi: done

pkg/dnsmasq/const.go
stuggi: done

config/default/manager_default_images.yaml
```diff
@@ -17,3 +17,6 @@ spec:
           value: quay.io/podified-antelope-centos9/openstack-memcached:current-podified
         - name: INFRA_REDIS_IMAGE_URL_DEFAULT
           value: registry.redhat.io/rhel9/redis-6:latest
+        # TODO create its own container image, instead of using neutron one
+        - name: INFRA_DNSMASQ_IMAGE_URL_DEFAULT
+          value: quay.io/tripleozedcentos9/openstack-neutron-server:current-tripleo
```
abays, line 22: Should we use `quay.io/podified-antelope-centos9/openstack-neutron-server:current-podified` since is it available?

abays: I guess if we're just going to change this to a dedicated `dnsmasq` in the near future, this might be okay for now.  I just know they're trying to get rid of the `tripleozedcentos9` images.

stuggi: yes, just pushed a fix for that

controllers/network/dnsmasq_controller.go
```diff
@@ -0,0 +1,500 @@
...
+		CreateFunc: func(e event.CreateEvent) bool {
+			if _, ok := e.Object.GetLabels()[networkv1.DNSDataLabelSelectorKey]; !ok {
+				return false
+			}
+
+			return true
+		},
```
gibizer, line 246: Do we need to watch for Delete events to remove deleted DNS config data from the dnsmasq config?

controllers/network/dnsmasq_controller.go
```diff
@@ -0,0 +1,500 @@
...
+	// what we want. If that is true, the event will be processed by the reconciler.
+	p := predicate.Funcs{
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			// The object doesn't contain label networkv1.DNSDataLabelSelectorKey, so the event will be
+			// ignored.
+			if _, ok := e.ObjectOld.GetLabels()[networkv1.DNSDataLabelSelectorKey]; !ok {
+				return false
+			}
+
+			return e.ObjectOld != e.ObjectNew
```
gibizer, line 238: I'm wondering if this does a proper per field value equality check or just compare pointer addresses of ObjectOld and ObjectNew.

controllers/network/dnsmasq_controller.go
```diff
@@ -0,0 +1,500 @@
...
+	return r.reconcileNormal(ctx, instance, helper)
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *DNSMasqReconciler) SetupWithManager(mgr ctrl.Manager) error {
+
+	dnsmasqFN := handler.EnqueueRequestsFromMapFunc(func(o client.Object) []reconcile.Request {
+		result := []reconcile.Request{}
+
+		// Get a list of all DNSMasq to trigger reconcile for the one in the same namespace
```
gibizer, line 201: Please add a bit more explanation. Something like:
// For each ConfigMap create / update event get the list of all ...

controllers/network/service_controller.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+
+	return ctrl.Result{}, nil
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *ServiceReconciler) SetupWithManager(mgr ctrl.Manager) error {
+
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&corev1.Service{}).
```
gibizer, line 104: I guess this could be also limited to services with certain labels / annotations (e.g. `svc.Annotations[networkv1.AnnotationHostnameKey]`) so that we only get triggered for a subset of services and not all services in the whole k8s cluster. (Can be done in a follow up)

controllers/network/service_controller.go
```diff
@@ -0,0 +1,198 @@
...
+// the Service object against the actual cluster state, and then
+// perform operations to make the cluster state reflect the state specified by
+// the user.
+//
+// For more details, check Reconcile and its Result here:
+// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.13.0/pkg/reconcile
+func (r *ServiceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	_ = log.FromContext(ctx)
+
+	// Get a list of all DNSMasqs to trigger reconcile only for these namespaces
```
gibizer, line 65: This is a bit misleading. Below you limit the query to the single namespace the Service is in. I would say instead:
// Get a list of DNSMasq CRs which are in the same namespace the Service is in to add the Service to those DNSMasqs.

controllers/network/service_controller.go
```diff
@@ -0,0 +1,198 @@
...
+}
+
+// getServiceDNSData -
+func (r *ServiceReconciler) getServiceDNSData(
+	ctx context.Context,
+	namespace string,
+) (map[string]networkv1.DNSHost, error) {
+	svcDNSHosts := map[string]networkv1.DNSHost{}
+
+	// get all services from the namespace triggered the reconcile
```
gibizer, line 115: I guess we do this so that we can regenerate the whole DNSHost list for a namespace for each reconcile instead of trying to modify some specific items in DNSHost list based on the Service that is changed.  This is clearly a tradeoff between simpler code vs more efficient but more complex code. If this become a performance bottleneck then we can avoid listing all the Services in the namespace and just modify the DNSHost data for the given Service that is being reconciled here.

controllers/network/service_controller.go
```diff
@@ -0,0 +1,198 @@
...
+
+	listOpts := []client.ListOption{
+		client.InNamespace(req.Namespace),
+	}
+	if err := r.Client.List(ctx, dnsmasqs, listOpts...); err != nil {
+		return ctrl.Result{}, fmt.Errorf("Unable to retrieve DNSMasqList %w", err)
+	}
+
+	for _, dnsmasq := range dnsmasqs.Items {
+		dnsHosts, err := r.getServiceDNSData(ctx, dnsmasq.Namespace)
```
gibizer, line 76: This is a repeated call for each DNSMasq CR, but we know that each of those is in the same Namespace as you limited the DNSMasq query to the req.Namespace above. So it is enough to call getServiceDNSData once before the loop.

controllers/network/service_controller.go
```diff
@@ -0,0 +1,198 @@
...
+		sortedDNSHosts := []networkv1.DNSHost{}
+
+		keys := maps.Keys(dnsHosts)
+		sort.Strings(keys)
+
+		for _, key := range keys {
+			sortedDNSHosts = append(sortedDNSHosts, dnsHosts[key])
+		}
+
+		err = r.createOrPatchDNSData(ctx, dnsmasq.Namespace, &dnsmasq.ObjectMeta, dnsmasq.Spec.DNSDataLabelSelectorValue, sortedDNSHosts)
```
gibizer, line 91: I think you can pass the whole dnsmasq object you got from the List query above to createOrPatchDNSData so that the createOrPatchDNSData does not need to Get it again.

controllers/network/service_controller.go
```diff
@@ -0,0 +1,198 @@
...
+func (r *ServiceReconciler) createOrPatchDNSData(
+	ctx context.Context,
+	namespace string,
+	owner metav1.Object,
+	dnsDataLabelSelectorValue string,
+	svcDNSHosts []networkv1.DNSHost,
+) error {
+	// Fetch the DNSMasq instance
+	instance := &networkv1.DNSMasq{}
+	err := r.Client.Get(ctx, types.NamespacedName{Name: owner.GetName(), Namespace: owner.GetNamespace()}, instance)
```
gibizer, line 164: You can get the DNSMasq instance from the caller instead.

gibizer (changes requested):
I haven't looked at the test code yet. I will do that next. But it will take time so I publish my comments on the code now.

config/samples/network_v1beta1_dnsdata.yaml
```diff
@@ -0,0 +1,45 @@
+apiVersion: network.openstack.org/v1beta1
+kind: DNSData
+metadata:
+  name: allhosts
+spec:
```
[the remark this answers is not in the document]
stuggi: right, services which get exposed via metallb loadbalancer type, they auto set the required annotation that service_controller will auto add them to the svc dnsdata  cr. for rabbitmq its automated via openstack-operator as in https://github.com/openstack-k8s-operators/openstack-operator/pull/297 . the dataplane operator would have to created dnsdata CRs, probably per role.

apis/network/v1beta1/webhook_suite_test.go
```diff
@@ -0,0 +1,132 @@
...
+var ctx context.Context
+var cancel context.CancelFunc
+
+func TestAPIs(t *testing.T) {
+	RegisterFailHandler(Fail)
+
+	RunSpecs(t, "Webhook Suite")
+}
+
+var _ = BeforeSuite(func() {
```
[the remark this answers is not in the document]
stuggi: ack

controllers/network/dnsdata_controller.go
```diff
@@ -0,0 +1,228 @@
...
+		r.Log,
+	)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	// Always patch the instance status when exiting this function so we can persist any changes.
+	defer func() {
+		// update the Ready condition based on the sub conditions
+		if allSubConditionIsTrue(instance.Status.Conditions) {
```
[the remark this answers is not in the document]
stuggi: right, missed to change that after I added it to lib-common. thanks for the pointer

controllers/network/dnsmasq_controller.go
```diff
@@ -0,0 +1,500 @@
...
+		instance.Status.Conditions.Set(condition.FalseCondition(
+			condition.DeploymentReadyCondition,
+			condition.RequestedReason,
+			condition.SeverityInfo,
+			condition.DeploymentReadyRunningMessage))
+		return ctrlResult, nil
+	}
+	instance.Status.ReadyCount = depl.GetDeployment().Status.ReadyReplicas
+
+	if instance.Status.ReadyCount > 0 {
```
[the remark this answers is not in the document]
stuggi: right, I think I just want to check if it matches the replicas from the spec as that is what we requested. I remember we discussed this a while back in the context of the cinder operator

controllers/network/dnsmasq_controller.go
```diff
@@ -0,0 +1,500 @@
...
+		return ctrl.Result{}, nil
+	}
+	// create Deployment - end
+
+	r.Log.Info("Reconciled Service successfully")
+	return ctrl.Result{}, nil
+}
+
+// hashOfInputHashes - calculates the overall hash of all our inputs
+func hashOfInputHashes(
```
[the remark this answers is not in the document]
stuggi: gnah I already did that in https://github.com/openstack-k8s-operators/lib-common/pull/251 . will update to use it

controllers/network/service_controller.go
stuggi: yes, but afaik you can not limit using For and a For is required, so can not just have a Watch()

controllers/network/service_controller.go
stuggi: correct, I had the same thinking, but since we don't have that many services in a namespace , I think its ok to get the list. When we move the operators to its own namespace we should only have the ones from the openstack services there.  what we could do is in addition the the annotation, we also add a dnsmasq label, so that we could limit the resulting list to only those.

pkg/dnsmasq/deployment.go
```diff
@@ -0,0 +1,192 @@
...
+
+		//
+		// https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/
+		//
+		livenessProbe.TCPSocket = &corev1.TCPSocketAction{
+			Port: intstr.IntOrString{Type: intstr.Int, IntVal: int32(DNSPort)},
+		}
+		readinessProbe.TCPSocket = &corev1.TCPSocketAction{
+			Port: intstr.IntOrString{Type: intstr.Int, IntVal: int32(DNSPort)},
+		}
```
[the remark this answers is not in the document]
stuggi: it listens on TCP and UDP. since you can not test UDP properly, right now I expect that when its up and listens on TCP, UDP works as well. We probably want to look if there is a good way to improve that.

apis/network/v1beta1/webhook_suite_test.go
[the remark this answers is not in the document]
stuggi: done

stuggi:
> I haven't looked at the test code yet. I will do that next. But it will take time so I publish my comments on the code now.

Thanks a lot @gibizer , I think I have addressed your comments.

controllers/network/dnsmasq_controller.go
stuggi: I need to double check this

test/functional/base_test.go
```diff
@@ -0,0 +1,269 @@
...
+		},
+		Data: map[string]string{
+			name.Name: "172.20.0.80 keystone-internal.openstack.svc",
+		},
+	}
+	Expect(k8sClient.Create(ctx, configMap)).Should(Succeed())
+	return configMap
+}
+
+func CreateService(name types.NamespacedName, spec map[string]interface{}) client.Object {
```
stuggi, line 162: should move this to lib-common in  a follow up

test/functional/base_test.go
```diff
@@ -0,0 +1,269 @@
...
+			},
+		},
+	}
+
+	Expect(k8sClient.Status().Update(ctx, svc)).To(Succeed())
+
+	return svc
+}
+
+func GetService(name types.NamespacedName) *corev1.Service {
```
stuggi, line 243: should move this to lib-common in  a follow up

test/functional/base_test.go
```diff
@@ -0,0 +1,269 @@
...
+	instance := &corev1.Service{}
+	Eventually(func(g Gomega) {
+		g.Expect(k8sClient.Get(ctx, name, instance)).Should(Succeed())
+	}, timeout, interval).Should(Succeed())
+	return instance
+}
+
+func DeleteService(name types.NamespacedName) {
```
stuggi, line 251: should move this to lib-common in  a follow up

test/functional/base_test.go
stuggi: https://github.com/openstack-k8s-operators/lib-common/pull/264

test/functional/base_test.go
stuggi: https://github.com/openstack-k8s-operators/lib-common/pull/264

test/functional/dnsdata_controller_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+				th.DeleteInstance(GetDNSData(dnsDataName))
+
+				Eventually(func() []corev1.ConfigMap {
+					return th.ListConfigMaps(dnsDataName.Name).Items
+				}, timeout, interval).Should(BeEmpty())
+			})
+		})
+	})
+})
```
gibizer, line 114: Nice test cases! Thanks for adding these.

test/functional/dnsmasq_controller_test.go
```diff
@@ -0,0 +1,211 @@
...
+	var dnsMasqRoleBindingName types.NamespacedName
+	var deploymentName types.NamespacedName
+	var namespace string
+
+	BeforeEach(func() {
+		// NOTE(gibi): We need to create a unique namespace for each test run
+		// as namespaces cannot be deleted in a locally running envtest. See
+		// https://book.kubebuilder.io/reference/envtest.html#namespace-usage-limitation
+		namespace = uuid.New().String()
+		th.CreateNamespace(namespace)
```
gibizer, line 48: you can move it to a Before each for the suite to avoid this repetition

test/functional/dnsmasq_controller_test.go
```diff
@@ -0,0 +1,211 @@
...
+			CreateDNSdataConfigMap(dnsDataCM)
+			DeferCleanup(th.DeleteConfigMap, dnsDataCM)
+			DeferCleanup(th.DeleteInstance, instance)
+		})
+
+		It("should have the Spec and Status fields initialized", func() {
+			instance := GetDNSMasq(dnsMasqName)
+			Expect(instance.Status.Hash).To(BeEmpty())
+			Expect(instance.Status.ReadyCount).To(Equal(int32(0)))
+			Expect(instance.Spec.ContainerImage).Should(Equal(containerImage))
```
gibizer, line 88: you can assert the list of conditions are inited as well

test/functional/dnsmasq_controller_test.go
```diff
@@ -0,0 +1,211 @@
...
+			)
+
+			deploymentName = types.NamespacedName{
+				Name:      fmt.Sprintf("dnsmasq-%s", dnsMasqName.Name),
+				Namespace: namespace,
+			}
+
+			time.Sleep(1000 * time.Millisecond)
+
+			depl := th.GetDeployment(deploymentName)
```
gibizer, line 177: In general you can use Eventually to assert async conditions. GetDeployment already uses Eventually to wait for the Deployment to exists. So if the only thing the sleep was needed is to wait for the Deployment to be created then you can simply remove the sleep and the Eventually block in GetDeployment will wait for it.

In my local testing this is the case, i.e. I removed the sleep and the test still passes. 

But if for some reason the reconcile would first create the Deployment and then later it would update it (e.g. setting the Replicas) then GetDeployment might return right after the Deployment is created but before it was updated in a later reconcile. If you have this situation then you can use an Eventually block like:
```golang
Eventually(func(g Gomega){
    depl := th.GetDeployment(deploymentName)
    g.Expect(int(*depl.Spec.Replicas)).To(Equal(1))
}, timeout, interval).Should(Succeed())
```
This Eventually could then waits for both conditions: i) the GetDeployment returns a deployment so it exsists, ii) the Replicas in that deployment in equal to 1.

test/functional/service_controller_test.go
```diff
@@ -0,0 +1,145 @@
...
+	var dnsMasqName types.NamespacedName
+	var svcDNSData types.NamespacedName
+	var namespace string
+
+	BeforeEach(func() {
+		// NOTE(gibi): We need to create a unique namespace for each test run
+		// as namespaces cannot be deleted in a locally running envtest. See
+		// https://book.kubebuilder.io/reference/envtest.html#namespace-usage-limitation
+		namespace = uuid.New().String()
+		th.CreateNamespace(namespace)
```
gibizer, line 42: can we moved to a common place

test/functional/service_controller_test.go
```diff
@@ -0,0 +1,145 @@
...
+			DeferCleanup(th.DeleteInstance, svc)
+			DeferCleanup(th.DeleteInstance, instance)
+		})
+
+		It("should have created a DNSData with hosts data", func() {
+			svc := GetService(serviceName)
+			Expect(svc).To(Not(BeNil()))
+
+			// seems a wait is required, otherwise it can fail?!?
+			time.Sleep(1000 * time.Millisecond)
```
gibizer, line 76: OK, I could reproduce an intermittent failure locally if I remove this sleep.

I needed to add couple of more log statement to get a clear view what is the sequence of events.

Successful sequence:
```
  2023-05-11T10:48:39.834+0200	INFO	---Test---	Creating	{"raw": {"apiVersion":"network.openstack.org/v1beta1","kind":"DNSMasq","metadata":{"name":"bbc8d956-06c9-47f4-bd0e-f14a68c65863","namespace":"40218426-5c50-4f36-9a1b-58583d72ee5f"},"spec":{"containerImage":"test-dnsmasq-container-image","dnsDataLabelSelectorValue":"dnsdata","externalEndpoints":[{"ipAddressPool":"ctlplane","loadBalancerIPs":["internal-lb-ip-1","internal-lb-ip-2"]}],"options":[{"key":"server","values":["1.1.1.1"]}]}}}
  2023-05-11T10:48:39.934+0200	INFO	Starting workers	{"controller": "service", "controllerGroup": "", "controllerKind": "Service", "worker count": 1}
  2023-05-11T10:48:39.934+0200	INFO	Starting workers	{"controller": "dnsdata", "controllerGroup": "network.openstack.org", "controllerKind": "DNSData", "worker count": 1}
  2023-05-11T10:48:39.934+0200	INFO	Starting workers	{"controller": "dnsmasq", "controllerGroup": "network.openstack.org", "controllerKind": "DNSMasq", "worker count": 1}
  2023-05-11T10:48:39.934+0200	INFO	Reconciling	{"controller": "service", "controllerGroup": "", "controllerKind": "Service", "Service": {"name":"kubernetes","namespace":"default"}, "namespace": "default", "name": "kubernetes", "reconcileID": "48164718-f7d5-4c47-987d-8039739dc7c7", "req": "default/kubernetes"}
  2023-05-11T10:48:41.860+0200	INFO	Reconciling	{"controller": "dnsmasq", "controllerGroup": "network.openstack.org", "controllerKind": "DNSMasq", "DNSMasq": {"name":"bbc8d956-06c9-47f4-bd0e-f14a68c65863","namespace":"40218426-5c50-4f36-9a1b-58583d72ee5f"}, "namespace": "40218426-5c50-4f36-9a1b-58583d72ee5f", "name": "bbc8d956-06c9-47f4-bd0e-f14a68c65863", "reconcileID": "99c69042-60d2-4e7c-bb65-a40712780d3f", "req": "40218426-5c50-4f36-9a1b-58583d72ee5f/bbc8d956-06c9-47f4-bd0e-f14a68c65863"}
  2023-05-11T10:48:41.863+0200	INFO	Reconciling	{"controller": "service", "controllerGroup": "", "controllerKind": "Service", "Service": {"name":"some-service","namespace":"40218426-5c50-4f36-9a1b-58583d72ee5f"}, "namespace": "40218426-5c50-4f36-9a1b-58583d72ee5f", "name": "some-service", "reconcileID": "6dfa18ba-3ee2-4d3d-aba1-e36307f2434f", "req": "40218426-5c50-4f36-9a1b-58583d72ee5f/some-service"}
  2023-05-11T10:48:41.863+0200	INFO	controllers.Service	creating / updating DNSData	{"DNSHosts": []}
  2023-05-11T10:48:41.864+0200	INFO	controllers.Service	bbc8d956-06c9-47f4-bd0e-f14a68c65863-svc - operation: created
  2023-05-11T10:48:41.864+0200	INFO	Reconciling	{"controller": "service", "controllerGroup": "", "controllerKind": "Service", "Service": {"name":"some-service","namespace":"40218426-5c50-4f36-9a1b-58583d72ee5f"}, "namespace": "40218426-5c50-4f36-9a1b-58583d72ee5f", "name": "some-service", "reconcileID": "c9f96a24-d9db-4810-9c5a-9b9d8978a161", "req": "40218426-5c50-4f36-9a1b-58583d72ee5f/some-service"}
  2023-05-11T10:48:41.864+0200	INFO	controllers.Service	creating / updating DNSData	{"DNSHosts": [{"ip":"172.20.0.80","hostnames":["some-service.40218426-5c50-4f36-9a1b-58583d72ee5f.svc"]}]}
  2023-05-11T10:48:41.864+0200	INFO	Reconciling	{"controller": "dnsdata", "controllerGroup": "network.openstack.org", "controllerKind": "DNSData", "DNSData": {"name":"bbc8d956-06c9-47f4-bd0e-f14a68c65863-svc","namespace":"40218426-5c50-4f36-9a1b-58583d72ee5f"}, "namespace": "40218426-5c50-4f36-9a1b-58583d72ee5f", "name": "bbc8d956-06c9-47f4-bd0e-f14a68c65863-svc", "reconcileID": "3a59912a-5d7c-4b00-83ec-9107d5b75246", "req": "40218426-5c50-4f36-9a1b-58583d72ee5f/bbc8d956-06c9-47f4-bd0e-f14a68c65863-svc"}
  2023-05-11T10:48:41.864+0200	INFO	Reconciling	{"controller": "dnsmasq", "controllerGroup": "network.openstack.org", "controllerKind": "DNSMasq", "DNSMasq": {"name":"bbc8d956-06c9-47f4-bd0e-f14a68c65863","namespace":"40218426-5c50-4f36-9a1b-58583d72ee5f"}, "namespace": "40218426-5c50-4f36-9a1b-58583d72ee5f", "name": "bbc8d956-06c9-47f4-bd0e-f14a68c65863", "reconcileID": "65248cb1-8f27-487b-b7fb-945558437e9b", "req": "40218426-5c50-4f36-9a1b-58583d72ee5f/bbc8d956-06c9-47f4-bd0e-f14a68c65863"}
  2023-05-11T10:48:41.865+0200	INFO	---Test---	Getting DNSData	{"name": "40218426-5c50-4f36-9a1b-58583d72ee5f/bbc8d956-06c9-47f4-bd0e-f14a68c65863-svc"}
```
Failed sequence:
```
  2023-05-11T10:50:15.193+0200	INFO	---Test---	Creating	{"raw": {"apiVersion":"network.openstack.org/v1beta1","kind":"DNSMasq","metadata":{"name":"10a49d4e-96dc-4797-a38f-2c481b9e7dbc","namespace":"c1f18965-176a-49e7-b5b0-e367959794a8"},"spec":{"containerImage":"test-dnsmasq-container-image","dnsDataLabelSelectorValue":"dnsdata","externalEndpoints":[{"ipAddressPool":"ctlplane","loadBalancerIPs":["internal-lb-ip-1","internal-lb-ip-2"]}],"options":[{"key":"server","values":["1.1.1.1"]}]}}}
  2023-05-11T10:50:15.291+0200	INFO	Starting workers	{"controller": "service", "controllerGroup": "", "controllerKind": "Service", "worker count": 1}
  2023-05-11T10:50:15.291+0200	INFO	Reconciling	{"controller": "service", "controllerGroup": "", "controllerKind": "Service", "Service": {"name":"kubernetes","namespace":"default"}, "namespace": "default", "name": "kubernetes", "reconcileID": "a13ecc56-696e-4362-8521-877a992b4842", "req": "default/kubernetes"}
  2023-05-11T10:50:15.291+0200	INFO	Starting workers	{"controller": "dnsmasq", "controllerGroup": "network.openstack.org", "controllerKind": "DNSMasq", "worker count": 1}
  2023-05-11T10:50:15.291+0200	INFO	Starting workers	{"controller": "dnsdata", "controllerGroup": "network.openstack.org", "controllerKind": "DNSData", "worker count": 1}
  2023-05-11T10:50:17.213+0200	INFO	Reconciling	{"controller": "dnsmasq", "controllerGroup": "network.openstack.org", "controllerKind": "DNSMasq", "DNSMasq": {"name":"10a49d4e-96dc-4797-a38f-2c481b9e7dbc","namespace":"c1f18965-176a-49e7-b5b0-e367959794a8"}, "namespace": "c1f18965-176a-49e7-b5b0-e367959794a8", "name": "10a49d4e-96dc-4797-a38f-2c481b9e7dbc", "reconcileID": "be374dd3-a29d-4860-a98d-fd1003eb48bd", "req": "c1f18965-176a-49e7-b5b0-e367959794a8/10a49d4e-96dc-4797-a38f-2c481b9e7dbc"}
  2023-05-11T10:50:17.220+0200	INFO	Reconciling	{"controller": "service", "controllerGroup": "", "controllerKind": "Service", "Service": {"name":"some-service","namespace":"c1f18965-176a-49e7-b5b0-e367959794a8"}, "namespace": "c1f18965-176a-49e7-b5b0-e367959794a8", "name": "some-service", "reconcileID": "1b799208-e96e-4f79-84b9-408c58fb10a9", "req": "c1f18965-176a-49e7-b5b0-e367959794a8/some-service"}
  2023-05-11T10:50:17.220+0200	INFO	controllers.Service	creating / updating DNSData	{"DNSHosts": []}
  2023-05-11T10:50:17.223+0200	INFO	controllers.Service	10a49d4e-96dc-4797-a38f-2c481b9e7dbc-svc - operation: created
  2023-05-11T10:50:17.223+0200	INFO	Reconciling	{"controller": "dnsdata", "controllerGroup": "network.openstack.org", "controllerKind": "DNSData", "DNSData": {"name":"10a49d4e-96dc-4797-a38f-2c481b9e7dbc-svc","namespace":"c1f18965-176a-49e7-b5b0-e367959794a8"}, "namespace": "c1f18965-176a-49e7-b5b0-e367959794a8", "name": "10a49d4e-96dc-4797-a38f-2c481b9e7dbc-svc", "reconcileID": "1f64feda-542f-4552-9d15-b35aeb0c2404", "req": "c1f18965-176a-49e7-b5b0-e367959794a8/10a49d4e-96dc-4797-a38f-2c481b9e7dbc-svc"}
  2023-05-11T10:50:17.223+0200	INFO	Reconciling	{"controller": "service", "controllerGroup": "", "controllerKind": "Service", "Service": {"name":"some-service","namespace":"c1f18965-176a-49e7-b5b0-e367959794a8"}, "namespace": "c1f18965-176a-49e7-b5b0-e367959794a8", "name": "some-service", "reconcileID": "72c3caee-aa97-40f7-9b04-21d9fd3eb334", "req": "c1f18965-176a-49e7-b5b0-e367959794a8/some-service"}
  2023-05-11T10:50:17.223+0200	INFO	controllers.Service	creating / updating DNSData	{"DNSHosts": [{"ip":"172.20.0.80","hostnames":["some-service.c1f18965-176a-49e7-b5b0-e367959794a8.svc"]}]}
  2023-05-11T10:50:17.224+0200	INFO	Reconciling	{"controller": "dnsmasq", "controllerGroup": "network.openstack.org", "controllerKind": "DNSMasq", "DNSMasq": {"name":"10a49d4e-96dc-4797-a38f-2c481b9e7dbc","namespace":"c1f18965-176a-49e7-b5b0-e367959794a8"}, "namespace": "c1f18965-176a-49e7-b5b0-e367959794a8", "name": "10a49d4e-96dc-4797-a38f-2c481b9e7dbc", "reconcileID": "096dc5ed-66f8-4384-a577-b87b1c67ef21", "req": "c1f18965-176a-49e7-b5b0-e367959794a8/10a49d4e-96dc-4797-a38f-2c481b9e7dbc"}
  2023-05-11T10:50:17.224+0200	INFO	---Test---	Getting DNSData	{"name": "c1f18965-176a-49e7-b5b0-e367959794a8/10a49d4e-96dc-4797-a38f-2c481b9e7dbc-svc"}
  2023-05-11T10:50:17.225+0200	INFO	---Test---	Got DNSData	{"data": {"metadata":{"name":"10a49d4e-96dc-4797-a38f-2c481b9e7dbc-svc","namespace":"c1f18965-176a-49e7-b5b0-e367959794a8","uid":"d97868b7-ef1d-4efc-a482-ab02766a2e59","resourceVersion":"225","generation":1,"creationTimestamp":"2023-05-11T08:50:17Z","ownerReferences":[{"apiVersion":"network.openstack.org/v1beta1","kind":"DNSMasq","name":"10a49d4e-96dc-4797-a38f-2c481b9e7dbc","uid":"d0e6aa0a-29cc-4d84-8d9b-8a4d7bb9da9c","controller":true,"blockOwnerDeletion":true}],"finalizers":["DNSData"],"managedFields":[{"manager":"functional.test","operation":"Update","apiVersion":"network.openstack.org/v1beta1","time":"2023-05-11T08:50:17Z","fieldsType":"FieldsV1","fieldsV1":{"f:metadata":{"f:finalizers":{".":{},"v:\"DNSData\"":{}},"f:ownerReferences":{".":{},"k:{\"uid\":\"d0e6aa0a-29cc-4d84-8d9b-8a4d7bb9da9c\"}":{}}},"f:spec":{".":{},"f:dnsDataLabelSelectorValue":{}}}}]},"spec":{"dnsDataLabelSelectorValue":"dnsdata"},"status":{}}}
  [FAILED] in [It] - /home/gibi/upstream/git/openstack-k8s-operators/infra-operator/test/functional/service_controller_test.go:81 @ 05/11/23 10:50:17.225
```

gibizer: So the DNSData goes through a two step reconciliation. First it is created with empty DNSHost then it is updated with a proper DNSHost. I did not dig deeper so I'm not sure it is bug in the controller or it is a normal behavior. But it explains that why the test fails sometimes. If the test reads the DNSData after it was created but before it was updated then it will see an empty DNSHost list. While if the test reads it after both create and update then it gets the populated list.


* a) If this is not the expected behavior of the controller then the controller should be fixed.
* b) If this is the expected behavior of the controller then you can wrap the L78-L83 into an Eventually block to *wait* for the DNSData to be fully populated.
  ```golang
			Eventually(func(g Gomega) {
				dnsdata := GetDNSData(svcDNSData)
				g.Expect(dnsdata).To(Not(BeNil()))
				g.Expect(dnsdata.Spec.DNSDataLabelSelectorValue).To(Equal("dnsdata"))
				g.Expect(dnsdata.Spec.Hosts).To(Not(BeNil()))
				g.Expect(dnsdata.Spec.Hosts[0].IP).To(Equal("172.20.0.80"))
				g.Expect(dnsdata.Spec.Hosts[0].Hostnames[0]).To(Equal(fmt.Sprintf("some-service.%s.svc", namespace)))
			}, timeout, interval).Should(Succeed())
```

**Review by gibizer — COMMENTED** [review]
I reached the point of you test question about the sleep and I have answers :) I haven't finished reviewing the whole test suite but I publish my comments so far to let you progress with this.

**Review by gibizer — COMMENTED** [review]
I reached the point of you test question about the sleep and I have answers :) I haven't finished reviewing the whole test suite but I publish my comments so far to let you progress with this.

File: test/functional/service_controller_test.go (hunk 35 above)
Comment (stuggi) [reply test/functional/service_controller_test.go:76]: Thanks for helping me on this. Not sure why I have not considered using Eventually here.With the current implementation its valid to have empty DNSHost. This is to make sure that when all services with annotations got removed we remove all entries from the DNSData CR. I'll update the PR with using Eventually

File: test/functional/base_test.go [hunk 36]
```diff
@@ -0,0 +1,269 @@
...
+		},
+		"spec": spec,
+	}
+
+	return th.CreateUnstructured(raw)
+}
+
+func GetDefaultDNSMasqSpec() map[string]interface{} {
+	spec := make(map[string]interface{})
+	spec["containerImage"] = "test-dnsmasq-container-image"
```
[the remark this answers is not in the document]
stuggi: makes sense

test/functional/base_test.go
```diff
@@ -0,0 +1,269 @@
...
+		},
+		Status: corev1.ServiceStatus{
+			LoadBalancer: corev1.LoadBalancerStatus{
+				Ingress: []corev1.LoadBalancerIngress{
+					{
+						IP: "172.20.0.80",
+					},
+				},
+			},
+		},
```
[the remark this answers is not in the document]
stuggi: nice thanks

test/functional/dnsdata_controller_test.go
```diff
@@ -0,0 +1,114 @@
...
+			}
+
+			DeferCleanup(th.DeleteInstance, instance)
+		})
+
+		It("should have the Spec and Status fields initialized", func() {
+			instance := GetDNSData(dnsDataName)
+			Expect(instance.Status.Hash).To(BeEmpty())
+			Expect(instance.Spec.DNSDataLabelSelectorValue).To(Equal("someselector"))
+			Expect(instance.Spec.Hosts).To(HaveLen(2))
```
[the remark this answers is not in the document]
stuggi: its to fast to check for e.g. ReadyCondition is false. at this point the ReadyCondition is already true

test/functional/dnsdata_controller_test.go
```diff
@@ -0,0 +1,114 @@
...
+		// NOTE(gibi): We need to create a unique namespace for each test run
+		// as namespaces cannot be deleted in a locally running envtest. See
+		// https://book.kubebuilder.io/reference/envtest.html#namespace-usage-limitation
+		namespace = uuid.New().String()
+		th.CreateNamespace(namespace)
+		// We still request the delete of the Namespace to properly cleanup if
+		// we run the test in an existing cluster.
+		DeferCleanup(th.DeleteNamespace, namespace)
+
+	})
```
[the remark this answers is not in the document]
stuggi: wouldn't it then create just one namespace for all tests?

PR #32, opened 2023-04-24
Dynamic serviceaccount/role/rolebinding creation

dprince:
This PR dyamically creates the above resources within the namespace of the requested (client,redis,memcached) instance(s).

The above resources are no longer created via OLM.

This allows the operator to deploy openstack service into any namespace within the OpenShift cluster.


pkg/memcached/statefulset.go
```diff
@@ -57,7 +57,7 @@ func StatefulSet(m *memcachedv1.Memcached) *appsv1.StatefulSet {
 					Labels: ls,
 				},
 				Spec: corev1.PodSpec{
-					ServiceAccountName: "mariadb-operator-mariadb",
```
dprince, line 60: This was a bug. An explicit dependency on the mariadb operator's service "service account"

abays (approved):
/lgtm

PR #36, opened 2023-05-15
memcached: Add memcached- prefix to stateful set name

kajinamit:
... to avoid name collision with the other operators. This also allows us to name pods deployed by the statefulset with memcached- prefix, which helps identifying for which service the pod is deployed.

### DIFF
--- a/pkg/memcached/statefulset.go
+++ b/pkg/memcached/statefulset.go
@@ -1,6 +1,8 @@
 package memcached
 
 import (
+	"fmt"
+
 	memcachedv1 "github.com/openstack-k8s-operators/infra-operator/apis/memcached/v1beta1"
 	labels "github.com/openstack-k8s-operators/lib-common/modules/common/labels"
 	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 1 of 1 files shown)


kajinamit:
Currently horizon-operaror creates a Memcached CR with name `horizon`, which creates pod and statefulset with only `horizon` in their names.

```
$ oc get statefulset
NAME              READY   AGE
horizon           1/1     8m17s
rabbitmq-server   1/1     4d1h
$ oc get po
NAME                                                              READY   STATUS             RESTARTS        AGE
...
horizon-0                                                         1/1     Running            0               8m19s
...
```

IMO this is quite confusing and can cause a problem later (in case horizon-operator tries to create its own horizon statefuleset).

pkg/memcached/statefulset.go
```diff
@@ -43,7 +43,7 @@ func StatefulSet(m *memcachedv1.Memcached) *appsv1.StatefulSet {
 
 	sfs := &appsv1.StatefulSet{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      m.Name,
+			Name:      "memcached-" + m.Name,
```
abays, line 46: Should we use `fmt.Sprintf()` here instead?

PR #38, opened 2023-05-15
memcached: Add memcached- prefix to stateful set name

kajinamit:
... to avoid name collision with the other operators. This also allows us to name pods deployed by the statefulset with memcached- prefix, which helps identifying for which service the pod is deployed.


kajinamit:
Seems the original PR ( https://github.com/openstack-k8s-operators/infra-operator/pull/36 ) was closed by branch renaming.

PR #45, opened 2023-05-24
Present memcached server list as part of CR status

kajinamit:
This introduces the two new fields to present list of memcached servers to Memcached CR. This allows each operator to check ready condition and list of memcached endpoints by only querying Memcached CR instance, instead of querying Pod list separately.


kajinamit:
Tested in my local env.
```
$ oc get memcacheds/memcached -o yaml
apiVersion: memcached.openstack.org/v1beta1
kind: Memcached
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"memcached.openstack.org/v1beta1","kind":"Memcached","metadata":{"annotations":{},"name":"memcached","namespace":"openstack"},"spec":{"replicas":1,"secret":"osp-secret","storageClass":"local-storage"}}
  creationTimestamp: "2023-05-25T01:34:47Z"
  generation: 1
  name: memcached
  namespace: openstack
  resourceVersion: "9146492"
  uid: 57c78c5a-886b-4b91-8fac-81b7c3b14f1d
spec:
  containerImage: quay.io/podified-antelope-centos9/openstack-memcached:current-podified
  replicas: 1
status:
  conditions:
  - lastTransitionTime: "2023-05-25T01:35:03Z"
    message: Setup complete
    reason: Ready
    status: "True"
    type: Ready
  - lastTransitionTime: "2023-05-25T01:35:03Z"
    message: Deployment completed
    reason: Ready
    status: "True"
    type: DeploymentReady
  - lastTransitionTime: "2023-05-25T01:34:52Z"
    message: Exposing service completed
    reason: Ready
    status: "True"
    type: ExposeServiceReady
  - lastTransitionTime: "2023-05-25T01:34:52Z"
    message: RoleBinding created
    reason: Ready
    status: "True"
    type: RoleBindingReady
  - lastTransitionTime: "2023-05-25T01:34:52Z"
    message: Role created
    reason: Ready
    status: "True"
    type: RoleReady
  - lastTransitionTime: "2023-05-25T01:35:03Z"
    message: ServerList completed
    reason: Ready
    status: "True"
    type: ServerListReady
  - lastTransitionTime: "2023-05-25T01:34:52Z"
    message: ServiceAccount created
    reason: Ready
    status: "True"
    type: ServiceAccountReady
  - lastTransitionTime: "2023-05-25T01:34:52Z"
    message: Service config create completed
    reason: Ready
    status: "True"
    type: ServiceConfigReady
  serverList:
  - memcached-0.memcached:11211
  serverListWithInet:
  - inet:[memcached-0.memcached]:11211
```

dciabrin (commented):
forewords: I think there is a competing concerns between how we expose the memcache endpoints and how oslo.cache consumes them. The latter is supposed to handle temporary disconnection/retries/black lists for endpoints which are not responding.

Now I'm not found of how the operator currently exposes the endpoints because it relies on querying available pods, rather that building up a list for "all the endpoints that will eventually be available". That means that openstack service might get an incomplete list of endpoint if the list is being generated while a pod is not started yet or is being stopped due to some temporary error.

I tend to think this list should be built statically based on the memcache spec, and the condition should be set to ready as soon as one memcache pod becomes available. Then oslo.cache should probably be responsible for handling endpoints availability.

(Note that my point above does not consider how openstack services should react to scaleup/scaledown of memcache endpoints in the memcache CR.)

Takashi could you elaborate on how you envision this list of pods will be consumed? Maybe the field is still a nice to have, but it should be generated out of the spec and not out of list of pods? Maybe you disagree with my point?

kajinamit:
> Takashi could you elaborate on how you envision this list of pods will be consumed? Maybe the field is still a nice to have, but it should be generated out of the spec and not out of list of pods? Maybe you disagree with my point?

You can find the example implementation to leverage these fields in https://github.com/kajinamit/horizon-operator/commit/8a05c84f3b9ad43255ec4b94d6eaca06aeaf7c36 . In short the plan is to let operators deploy memcached and obtain these list from status, then put the list items into config file.

We can probably create the list statically instead of querying pods, though we have to be aware that we have to simulate the logic inside Statefulset for it (the logic can be broken by any change in kore k8s though I think that unlikely happen).

I'll look for the way to implement these in more static manner but I still think we should have the common logic in memcached operator, because
 - We need the same logic in all operators deploying memcached
 - We still need to properly detect IP version used by memcached (unless we can detect this in a different way, instead of querying pod addresses)

kajinamit:
OK we can probably query service instead of pods to get ip version.

```
$ oc get services/horizon -o yaml
apiVersion: v1
kind: Service
...
spec:
  clusterIP: None
  clusterIPs:
  - None
  internalTrafficPolicy: Cluster
  ipFamilies:
  - IPv4
  ipFamilyPolicy: SingleStack
  ports:
  - name: memcached
    port: 11211
    protocol: TCP
    targetPort: 11211
  selector:
    app: memcached
  sessionAffinity: None
  type: ClusterIP
```

kajinamit:
> Hostname format is heavily dependent on how these pods are deployed, and memcached controller is maintaining that
One good example is the service name. How memcached controllers deploys service name according to instance name, and because of that if we deploy Memcached with name horizon then it creates a service with name: heat.

This is quite confusing and we ideally should have memcached- prefix but if we keep the logic to generate hostnames in each operators then we have to update all operators when we fix that.

apis/memcached/v1beta1/memcached_types.go
```diff
@@ -45,6 +45,12 @@ type MemcachedSpec struct {
 type MemcachedStatus struct {
 	// Conditions
 	Conditions condition.Conditions `json:"conditions,omitempty" optional:"true"`
+
+	// ServerList - List of memcached endpoints without inet(6) prefix
+	ServerList []string `json:"serverList,omitempty" optional:"true"`
+
+	// ServerListWithPrefix - List of memcached endpoints with inet(6) prefix
+	ServerListWithPrefix []string `json:"serverListWithInet,omitempty" optional:"true"`
```
gibizer, line 53: * Out of curiosity: Why we need to have two lists?
* The golang struct name `ServerListWithPrefix` could be synced with the name of the json field. (I prefer the json field name)

controllers/memcached/memcached_controller.go
```diff
@@ -256,3 +287,46 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		"memcached/name": instance.Name,
+	}
+	podList, err := oko_pod.GetPodListWithLabel(ctx, h, instance.Namespace, labelSelector)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	if len(podList.Items) == 0 {
+		return nil, nil, fmt.Errorf("No pods could be found")
+	}
```
gibizer, line 312: I'm not 100% sure this is an error. It could be a transient state after the statefulset is created but no pods are created yet. Note that if you return error from the reconciliation then a stack trace will be emitted to the log and the reconciliation request is requeued immediately. So this can lead to noisy logs.

controllers/memcached/memcached_controller.go
```diff
@@ -256,3 +287,46 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+
+	for _, pod := range podList.Items {
+		if pod.Spec.Hostname == "" || pod.Spec.Subdomain == "" || pod.Status.PodIP == "" {
+			return nil, nil, fmt.Errorf("Pod does not have the required Spec to construct endpoint")
+		}
```
gibizer: Similarly I would not treat this as an error in the reconciliation itself. This pod is probably not running yet, so you can simply ignore it here and process the rest of the pods.

apis/memcached/v1beta1/memcached_types.go
kajinamit: In OpenStack we have two implementations to use memcache. One is based on python-memcached while the other depends on pymemcache. pymemcache is much modern and does not require special treatment for IPv6, but python-memcached requires inet6 prefix in case the service uses IPv6 address.

We may probably just expose prefix and let clients compose the full server list but for that we also have to expose port. So I decided to compose the list items here in memcached controller.

I agree with the naming preference so I'll fix that.

controllers/memcached/memcached_controller.go
kajinamit: Alternatively we can let the controller wait for a while with inprogress status, though we don't expect we usually hit this because we run this logic only when at least one replica is available.

kajinamit: Fixed in the latest version

controllers/memcached/memcached_controller.go
kajinamit: Fixed in the latest version

apis/memcached/v1beta1/memcached_types.go
gibizer: I see. Thanks for the explanation. If we need the prefix for hte clients then I think we can keep the two lists. I would like to ask you to summarize the above reasoning in a code comment so that we will know in the future why we have a the duplication

kajinamit: I've added a comment in the logic.

kajinamit:
So looking into k8s blueprint and actual behavior in my env, I concluded that Service.spec.IPfamilies[0] can be used to determine which IP version is used for a service. A remaining problem is that we need additional handling for dual-stack use case but we've never support it in TripleO and we have to address more problems (eg we have to implement the way that all services use preferred IP version)  so I think we can leave the problem for a future TODO.

I've submitted https://github.com/openstack-k8s-operators/lib-common/pull/282 so that we have that information exposed by lib-common then I'll update it once that PR is merged.

gibizer (commented):
My comments are answered. So it looks OK to me.

abays (approved):
/lgtm

PR #46, opened 2023-05-25
Memcached: Present number of pods in status

kajinamit:
This adds ReadyCount field in status, so that users can understand number of pods properly deployed. This field is supported by most of CRs and adding this to Memcached CR makes the interface more consistent across all CRs.


abays (approved):
/lgtm

PR #51, opened 2023-05-30
Reset Ready condition when update failed

kajinamit:
This ensures the Ready Condition is reset when CR becomes once ready but something fails during reconfiguration.


abays (approved):
/lgtm

PR #52, opened 2023-06-05
Fix app label for memcached resources

kajinamit:
We have to use different labels for different CR instance because we deploy one service resource per CR instance.


abays (approved):
/lgtm

PR #59, opened 2023-06-19
Match vlan ids

rabi:
Use the same vlan ids we configure with nncp and use in edpm nodes.

### DIFF
--- a/config/samples/network_v1beta1_netconfig.yaml
+++ b/config/samples/network_v1beta1_netconfig.yaml
@@ -40,20 +40,20 @@ spec:
       - end: 172.18.0.250
         start: 172.18.0.10
       cidr: 172.18.0.0/24
-      vlan: 30
+      vlan: 21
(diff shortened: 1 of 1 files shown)


config/samples/network_v1beta1_netconfig.yaml
```diff
@@ -40,20 +40,20 @@ spec:
...
+      vlan: 21
   - name: StorageMgmt
     subnets:
     - name: subnet1
       allocationRanges:
       - end: 172.19.0.250
         start: 172.19.0.10
       cidr: 172.19.0.0/24
-      vlan: 40
+      vlan: 22
```
stuggi, line 51: right now tenant has `22` and the `172.19.0.0/24` cidr in install_yamls https://github.com/openstack-k8s-operators/install_yamls/blob/main/scripts/gen-nncp.sh#L91 .

stuggi (approved):
/lgtm

PR #60, opened 2023-06-20
Fix hostname in server list fields

kajinamit:
Pod DNS is supposed to look like `<stateful set name>-<index>.<service name>`

This fixes the wrong hostname format used to generate server list status fields.

This also removes redundant `memcached` from resource names. Now individual operators are supposed add it as CR name suffix, thus adding it again in memcached controller causes redundant names.


stuggi (approved):
/lgtm

PR #61, opened 2023-06-20
[IPAM] extend IPSet and NetConfig validation and adds Immutable feature

stuggi:
**A)** Adds further validations for IPSet and NetConfig.
NetConfig:

IPSet:
* Create
  - Validates that there is a NetConfig in the namespace

* Create and Update
  - networks are uniq in the list
  - networks and subnets exist in netcfg
  - FixedIP is a valid IP address
  - FixedIP has correct IP version of subnet
  - FixedIP is in the subnet cidr
  - Route is only specified on a single network

* Update
  - if a previous requested network is still in the list
  - if subnet changed within a network
  - if fixedIP changed
  - if defaultRoute changed

NetConfig:
* Create
  - validates if there is already a NetConfig in the namespace

* Update
   **Note**: those checks will be skipped if there are no IPSets in the namespace. in that case it is save to customize the NetConfig as needed.
  - validate if a network was removed
  - subnet is still there
  - cidr changed
  - Vlan changed
  - gateway changed
 
* Delete
  - validates that there is no IPSet in the namespace

**B)** Adds immutable parameter to IPSet

When Spec.Immutable == true on the existing object spec, the IPSet validation webhook on Update denies the spec change. Changes to the status and metadata are still possible as it is required for the operator to work properly.

Jira: [OSP-19996](https://issues.redhat.com//browse/OSP-19996)

### DIFF
--- a/apis/network/v1beta1/common_webhook.go
+++ b/apis/network/v1beta1/common_webhook.go
@@ -1,0 +1,80 @@
+package v1beta1
+
+// Because webhooks *MUST* exist within the api/<version> package, we need to place common
+// functions here that might be used across different Kinds' webhooks
+
+import (
+	"context"
+	"fmt"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	goClient "sigs.k8s.io/controller-runtime/pkg/client"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// Client needed for API calls (manager's client, set by first SetupWebhookWithManager() call
+// to any particular webhook)
+var webhookClient goClient.Client
+
+const (
+	errNotIPAddr              = "not an IP address"
+	errInvalidCidr            = "IP address prefix (CIDR) %s"
+	errNotInCidr              = "address not in IP address prefix (CIDR) %s"
+	errMixedAddressFamily     = "cannot mix IPv4 and IPv6"
+	errInvalidRange           = "Start address: %s > End address %s"
+	errDupeNetworkName        = "network name %s already in use at %s, must be uniq"
+	errDupeCIDR               = "CIDR %s already in use at %s"
+	errInvalidDNSDomain       = "DNSDoman name %s is not valid"
+	errDupeDNSDomain          = "DNSDoman name %s already in use at %s, must be uniq"
+	errNetworkNotFound        = "network %s not in NetConfig"
+	errSubnetNotInNetwork     = "subnet %s not in network %s of the NetConfig"
+	errNetworkChanged         = "network %s removed"
+	errSubnetChanged          = "subnet within a network must not change (%s/%s)"
+	errSubnetParameterChanged = "%s of a subnet must not change - %v"
+	errFixedIPChanged         = "fixedIP must not change"
+	errDefaultRouteChanged    = "defaultRoute must not change"
+	errMultiDefaultRoute      = "%s defaultRoute can only be requested on a singe network"
+)
+
+func getNetConfig(
+	c client.Client,
+	obj metav1.Object,
+) (*NetConfig, error) {
+	// check if NetConfig is available
+	opts := &client.ListOptions{
+		Namespace: obj.GetNamespace(),
+	}
+
+	netcfgs := &NetConfigList{}
+	err := webhookClient.List(context.TODO(), netcfgs, opts)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(netcfgs.Items) == 0 {
+		return nil, nil
+	} else if len(netcfgs.Items) > 1 {
+		return nil, fmt.Errorf("more then one NetConfig found in namespace %s, there must be only one", obj.GetNamespace())
+	}
+
+	return &netcfgs.Items[0], nil
+}
+
+func getIPSets(
+	c client.Client,
+	obj metav1.Object,
+) (*IPSetList, error) {
+	// check if IPSet is available
+	opts := &client.ListOptions{
+		Namespace: obj.GetNamespace(),
+	}
+
+	ipsets := &IPSetList{}
+	err := webhookClient.List(context.TODO(), ipsets, opts)
+	if err != nil {
+		return nil, err
+	}
+
+	return ipsets, nil
+}
--- a/apis/network/v1beta1/ipset_webhook.go
+++ b/apis/network/v1beta1/ipset_webhook.go
@@ -24,6 +24,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	k8snet "k8s.io/utils/net"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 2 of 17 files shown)


apis/network/v1beta1/common_webhook.go
```diff
@@ -0,0 +1,78 @@
...
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	goClient "sigs.k8s.io/controller-runtime/pkg/client"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// Client needed for API calls (manager's client, set by first SetupWebhookWithManager() call
+// to any particular webhook)
+var webhookClient goClient.Client
```
abays, line 18: It looks like only `ipset_webhook.go` ever sets the `webhookClient` right now.  Just wondering if you intended to add it to all webhooks given what the comment says.

apis/network/v1beta1/ipset_webhook.go
```diff
@@ -74,3 +133,124 @@ func (r *IPSet) ValidateDelete() error {
...
+						return allErrs
+					}
+
+					path := path.Child("fixedIP")
+					if err := valiateAddress(*_net.FixedIP, ipPrefix, path); err != nil {
+						allErrs = append(allErrs, err...)
+					}
+				}
+
+				// check that there are not multiple have the defaultRoute flag
```
hjensas, line 191: We need to allow one for IPv4 and one for IPv6. Not critical for now, can be a follow up.

apis/network/v1beta1/common_webhook.go
stuggi: probably not to all, but I'll add webhook to dns which will also required it. thats why I added it here

stuggi:
```
step infra-operator-build-deploy failed: failed to created cluster claim 9b0fe0ce-4875-4bbc-86c4-b68d34ef4604 in namespace openstack-k8s-operators-cluster-pool: etcdserver: request timed out
```

/retest

apis/network/v1beta1/ipset_webhook.go
```diff
@@ -74,3 +133,124 @@ func (r *IPSet) ValidateDelete() error {
...
+		netIdx := slices.IndexFunc(networks, f)
+		if netIdx < 0 {
+			// the network was removed
+			allErrs = append(allErrs, field.Invalid(path.Child("name"), _net.Name, fmt.Sprintf(errNetworkChanged, _net.Name)))
+			return allErrs
+		}
+
+		// validate if subnet changed within a network
+		if !equality.Semantic.DeepEqual(_net.SubnetName, networks[netIdx].SubnetName) {
+			allErrs = append(allErrs, field.Invalid(path.Child("subnetName"), _net.Name, fmt.Sprintf(errSubnetChanged, *&oldNetworks[_netIdx].SubnetName, *&networks[_netIdx].SubnetName)))
```
abays, line 243: Is there a reason that we use `*&` as opposed to just excluding that?

apis/network/v1beta1/netconfig_webhook.go
```diff
@@ -144,6 +172,79 @@ func valiateNetworks(
...
+				if _subnet.Gateway != nil {
+					gateway = *_subnet.Gateway
+				}
+				allErrs = append(allErrs, field.Invalid(path.Child("vlan"), _net.Name, fmt.Sprintf(errSubnetParameterChanged, "vlan", gateway)))
+			}
+
+		}
+		if len(allErrs) > 0 {
+			return allErrs
+		}
```
abays, line 242: Do we want to `return` here if there are errors for this network?  Just curious if it would make sense to check the other networks and report their potential errors as well.

apis/network/v1beta1/ipset_webhook.go
stuggi: no, fixing it

apis/network/v1beta1/netconfig_webhook.go
stuggi: I'll change it to have the same as on create. I thought it might make sense to report issues per network as it might be easier to understand. But now see that its different from create

apis/network/v1beta1/ipset_webhook.go
stuggi: done

apis/network/v1beta1/netconfig_webhook.go
stuggi: done

apis/network/v1beta1/ipset_webhook.go
stuggi: I have added this in the latest version I pushed

apis/network/v1beta1/ipset_webhook.go
```diff
@@ -74,3 +134,134 @@ func (r *IPSet) ValidateDelete() error {
...
+	return allErrs
+}
+
+// valiateIPSetChanged
+// - if a previous requested network is still in the list
+// - if subnet changed within a network
+// - if fixedIP changed
+// - if defaultRoute changed
+func valiateIPSetChanged(
+	networks []IPSetNetwork,
```
rabi, line 232: Can we not mark IPSet CRD immutable with the following kubebuilder comment and won't need to implement things in webhook for update? 

// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"

apis/network/v1beta1/netconfig_webhook.go
```diff
@@ -144,6 +172,76 @@ func valiateNetworks(
 	return allErrs
 }
 
+// valiateNetworksChanged
+// - validate if a network was removed
+// - subnet is still there
+// - cidr changed
+// - Vlan changed
+// - gateway changed
+func valiateNetworksChanged(
```
rabi, line 181: Maybe we can do the same XValidation kubebuilder comment for some fields?

abays (commented):
+1 from me.  Will approve if someone else seconds.

apis/network/v1beta1/ipset_webhook.go
stuggi: haven't check `XValidation`, will look into that. in general how about adding a new network to a compute node or a service? shouldn't this be possible?

how about this approach,  we add an immutable flag to the ipset and if set block any update, if not its up to the client? for a service on OCP which uses ipset it should be ok to update it, as it just gets restarted with the new settings.

rabi: > how about adding a new network to a compute node?

I've not seen cases of customers adding new networks on existing compute nodes, though it could be a valid use-case.

> how about this approach, we add an immutable flag to the ipset and if set block any update, if not its up to the client?

With the amount of support cases that we've seen where customer was not using fixed/predictable IP and IP changes causing chaos, I think making it immutable/adding an immutable flag would probably be better. I guess we would also need to add finalizers on IPSets from role/node. I did not add them atm because there is not reconcile delete logic for node/role atm for scale-down.

stuggi: ack, let me add that immutable flag, which the client then can set and we can set it from the dataplane-operator.

stuggi: @rabi @abays this is now added in a separate commit. Might make review easier. `Spec.Immutable` on the IPSet can be set to block spec updates.
I was wondering if it make sense to still be able to flip the Immutable for future fixing of envs. means if Immutable set you can not update the rest of the spec, but if required you have to explicitly update the object flip the Immutable and then could do changes. but can do that in a follow up if you think that  makes sense.

stuggi:
rebased

rabi (commented):
+1, lgtm

apis/network/v1beta1/ipset_webhook.go
hjensas: I've seen adding new networks a few times, for example to enable use of some new NFS or iSCSI storage backend they need to add VLAN/interface. I like the `Immutable` approach - forcing an additional step to allow changes makes sense and sends a good "be careful" signal.

hjensas (approved):
looks good, further improvements can be done in follow ups.

PR #62, opened 2023-06-20
Add Loadbalancer IPs to dnsmasq status

rabi:
This would help us to get it from the CR and use it in configuring dataplane. One we've fqdns we can use them.

Also fixes a few linting issues.


apis/bases/network.openstack.org_dnsmasqs.yaml
```diff
@@ -185,6 +185,11 @@ spec:
                   - type
                   type: object
                 type: array
+              dnsAddresses:
+                description: DNSServer Addresses
+                items:
+                  type: string
+                type: array
```
stuggi, line 192: thinking about if this needs to be an array. the IP address for dns is a single metallb vip. but it won't hurt to use an array.

stuggi: right we use also a list when requesting IPs via the CRD.

controllers/network/dnsmasq_controller.go
```diff
@@ -439,6 +439,8 @@ func (r *DNSMasqReconciler) reconcileNormal(ctx context.Context, instance *netwo
 		return ctrl.Result{}, nil
 	}
 	// create Deployment - end
+	//Update status with LoadBalancerIPs
+	instance.Status.DNSAddresses = instance.Spec.ExternalEndpoints[0].LoadBalancerIPs
```
stuggi, line 443: this will only work when specific LoadBalancerIPs are requested via the spec. If not, one from the metallb pool is assigned and its required lookup them from the LoadBalancer service. can use GetServiceWithName() [1] or GetServicesListWithLabel from the service pkg to get the service and then GetExternalIPs()  [2] for the IPs.

[1] https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/common/service/service.go#L250
[2] https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/common/service/service.go#L61

rabi: I can change, but I thought we would always use a fixed ip for dns service, unless we use some fqdn name for the dynamic lb ips?

stuggi: don't think we require it to be a fixed ip, you could. The dns service is not required for the ctlplane to be deployed and its the first being deployed https://github.com/openstack-k8s-operators/openstack-operator/blob/281f62ce0f40c3c82abb40bb422e6bdb92bcda46/controllers/core/openstackcontrolplane_controller.go#L149 . So in the workflow its there when the first edpm gets setup. Or do I miss a scenario?

stuggi (approved):
/lgtm

PR #63, opened 2023-06-20
[IPAM] add DNSDomain to the network and subnet definition

stuggi:
This allows to specify the dns domain name of the network and optional on the subnet. If none is specified on the subnet, the required value from the network is being added to the ipset status when register an ip address. If the DNSDomain is specified on the subnet it will be used.

Jira: [OSP-19996](https://issues.redhat.com//browse/OSP-19996)


apis/network/v1beta1/netconfig_types.go
```diff
@@ -52,6 +52,10 @@ type Subnet struct {
 	// Cidr the cidr to use for this network
 	Cidr string `json:"cidr"`
 
+	// +kubebuilder:validation:Required
+	// DNSDomain name of the subnet
+	DNSDomain string `json:"dnsDomain"`
```
stuggi, line 57: or should it be ok to just have a dnsdomain per network?

rabi: I think we had one per network in TripleO[1]. I guess it's not technically incorrect to have separate one per subnet(?).

[1] https://docs.openstack.org/project-deploy-guide/tripleo-docs/latest/features/custom_networks.html#network-data-yaml-options

hjensas: I think technically a dnsdomain don't have any actual connection to an IP subnet or network.
Assuming the different dnsdomains are for different service networks, i.e we map services to different networks (`ServiceNetMap` in  TripleO). In TripleO the idea with networks + subnets is to be able to map a service to a single network and still allow L3 (spine-leaf/routed). i.e unless we are changing the pattern to map service to network having `dnsdomain` per network is what we want?

stuggi: > I think technically a dnsdomain don't have any actual connection to an IP subnet or network. Assuming the different dnsdomains are for different service networks, i.e we map services to different networks (`ServiceNetMap` in TripleO).

right dnsdomain is not connected to subnets/network. there was a discussion from where a dataplanenode could get the domain/s on the networks to also register in dns. There was a comment that one might want to model dns entries like `<hostname>.<rack>.<room>.<site>.<company>.<tld>`, which could work with the proposal if you create a subnet per rack and have the domain set there.

Could also have it on the network and on the subnet as optional to be able to overwrite the one from the network.

> In TripleO the idea with networks + subnets is to be able to map a service to a single network and still allow L3 (spine-leaf/routed). i.e unless we are changing the pattern to map service to network having `dnsdomain` per network is what we want?
I am not sure if we require this service mapping <-> network mapping like we have in ooo. You expose the service via the route (maybe gateway api in future), the metallb service (addresspool) or the NAD. The NAD/addresspool need to be in any of the networks from NetConfig. The api services running on OCP get registered for the public endpoint with the route dns name, for the internal endpoint, exposed to an isolated network using `<service>-internal.openstack.svc` dns name. As we can not register multiple dns names and this allows the services on OCP to resolve it to the service ip and the edpm nodes via our dns server to the metallb vip. (https://github.com/openstack-k8s-operators/docs/blob/main/networking.md#ingress)

We probably require an annotation on the NAD to map it to a network, or a subnet so that we can register (if required) services which are exposed via multus in dns. This is probably a similar service mapping to a network?

As mentioned above, could also have the dnsdomain on the network and on the subnet as optional to be able to overwrite the one from the network. what is your take on this?

hjensas: > As mentioned above, could also have the dnsdomain on the network and on the subnet as optional to be able to overwrite the one from the network. what is your take on this?

I like that idea from UX pov, less typing for users. 
Those doing adoption and are likely to stay on the `<network>.<....>.<tld>` model used in ooo, so I guess there will be a lot of people just needing the ability to set it at nework level.

stuggi: > > As mentioned above, could also have the dnsdomain on the network and on the subnet as optional to be able to overwrite the one from the network. what is your take on this?
> 
> I like that idea from UX pov, less typing for users. Those doing adoption and are likely to stay on the `<network>.<....>.<tld>` model used in ooo, so I guess there will be a lot of people just needing the ability to set it at nework level.

I'll update the PR to add it to the network and make it optional for the subnet.

stuggi: I have updated the PR

hjensas (approved):
LGTM

PR #66, opened 2023-06-23
Add edpm jobs to infra-operator

rabi:
Now dataplane uses infra-operator for IPAM stuff. Let's avoid breaking the EDPM deployment with changes here.

Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/328

### DIFF
--- a/zuul.d/jobs.yaml
+++ b/zuul.d/jobs.yaml
@@ -1,0 +1,36 @@
+---
+- job:
+    name: infra-operator-content-provider
+    parent: content-provider-base
+    vars:
+      cifmw_operator_build_org: openstack-k8s-operators
+      cifmw_operator_build_operators:
+        - name: "openstack-operator"
+          src: "~/src/github.com/{{ cifmw_operator_build_org }}/openstack-operator"
+          image_base: infra
+    irrelevant-files: &irrev
+      - tests/kuttl
+      - docs
+      - containers/ci
+      - .github/workflows
+      - .ci-operator.yaml
+      - .dockerignore
+      - .gitignore
+      - .golangci.yaml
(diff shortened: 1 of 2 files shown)


zuul.d/jobs.yaml
```diff
@@ -0,0 +1,38 @@
...
+      - PROJECT
+      - .*/*.md
+      - kuttl-test.yaml
+      - renovate.json
+
+- job:
+    name: infra-operator-crc-podified-edpm-deployment
+    parent: cifmw-crc-podified-edpm-deployment
+    vars:
+      bmo_setup: true
```
raukadah, line 32: It is no longer needed, we can drop it.

raukadah (changes requested):

zuul.d/jobs.yaml
rabi: Why was that var added in the first place to crc jobs? It's by default true.

Sandeepyadav93:
/lgtm

Thanks Rabi

zuul.d/jobs.yaml
raukadah: this var was used with the older [crc parent](https://github.com/rdo-infra/review.rdoproject.org-config/blob/master/zuul.d/_jobs-crc.yaml#L4) and consumed [here](https://github.com/rdo-infra/review.rdoproject.org-config/blob/master/playbooks/crc/openstack.yaml#L39). We will be cleaning up soon.

raukadah:
recheck

Sandeepyadav93:
recheck

Sandeepyadav93:
recheck

Sandeepyadav93:
recheck

PR #76, opened 2023-06-28
Remove openstack from workflow name

Sandeepyadav93:
The image name/repo name for infra operator doesn't have openstack in its name. Removing openstack name so that we can sort github status list correctly [1].

[1] https://github.com/gibizer/openstack-k8s-status/tree/main


abays (approved):
/lgtm

PR #79, opened 2023-07-09
Merge test file directory

kajinamit:
This repository currently have the two directories to maintain test resources('test' and 'tests') but this is confusing. Let's merge these to into a single one.


kajinamit:
recheck

stuggi:
this PR is not merging the directories, just doing a change in the Makefile, is that expected?

kajinamit:
> this PR is not merging the directories, just doing a change in the Makefile, is that expected?

Hmm it's wired. It seems `git mv` was not properly included. Let me fix it now ...

stuggi:
in infra-operator, heat-operator its now `/tests` directory, in placement, node it is `/test` . might want to align there as well

stuggi (approved):
/lgtm

kajinamit:
> in infra-operator, heat-operator its now `/tests` directory, in placement, nova it is `/test` (not checked all others). might want to align there as well

Yeah. For now I'm renaming the directory in placement-operator by https://github.com/openstack-k8s-operators/placement-operator/pull/24 (it's a good timing because we can't directly rename kuttl directory once we add the reference to install_yamls Makefile).

kajinamit:
recheck

PR #82, opened 2023-07-11
Update test helper references after reorg

gibizer:
Depends-On: openstack-k8s-operators/lib-common#301


stuggi (commented):
also need an update to https://github.com/openstack-k8s-operators/infra-operator/blob/main/tests/functional/suite_test.go#L46

PR #84, opened 2023-07-12
Add a github action that checks for go.mod replace

kajinamit:
To catch unwanted local go.mod replace lines to be merged this patch adds a check for it.


stuggi (approved):
/lgtm

PR #88, opened 2023-07-16
Append default route in routes

rabi:
os-net-config used networkconfig templates expect default route appended to other ctlplane static routes.


stuggi (approved):
/lgtm

PR #93, opened 2023-07-19
Remove leftover of OpenStackClient CR

kajinamit:
We've removed OpenStackClient CR from infra-operator so now this operator does not depend on keystone-operator directly.


renovate.json
```diff
@@ -20,7 +20,7 @@
     {
       "groupName": "openstack-k8s-operators",
       "matchPackagePrefixes": ["github.com/openstack-k8s-operators"],
-      "excludePackageNames": ["github.com/openstack-k8s-operators/keystone-operator/api"],
+      "excludePackageNames": ["github.com/openstack-k8s-operators/infra-operator/api"],
```
abays, line 23: I'm not sure what this line should be doing.

1. Some operators have this similar `excludePackageNames` [1] but some don't [2].
2. If it should have it, `s/api/apis`

[1] https://github.com/openstack-k8s-operators/keystone-operator/blob/d341e7f8faa01401d41a74eaeaba5549bcbbbbe7/renovate.json#L23
[2] https://github.com/openstack-k8s-operators/glance-operator/blob/eaecd68d2867ae7e0fbf7df903ee536df064045b/renovate.json#L23-L25

kajinamit: 1. I think having this is correct, because we don't want to renovate to bump reference to own.
2. That was the typo. I fixed it.

Note that we originally have `infra-operator/apis` but it was updated by https://github.com/openstack-k8s-operators/infra-operator/pull/91 unintentionally. Probably @stuggi can double check it.

stuggi: right this was a mistake to change it to keystone-operator.

stuggi (approved):
/lgtm

PR #104, opened 2023-08-01
Bump k8s version in envtest

kajinamit:
... to the one consistent with the version in dependencies.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,7 +50,7 @@ endif
 DEFAULT_IMG ?= quay.io/openstack-k8s-operators/infra-operator:latest
 IMG ?= $(DEFAULT_IMG)
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
-ENVTEST_K8S_VERSION = 1.25.0
+ENVTEST_K8S_VERSION = 1.26
 
 # Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
 ifeq (,$(shell go env GOBIN))


Makefile
```diff
@@ -50,7 +50,7 @@ endif
 DEFAULT_IMG ?= quay.io/openstack-k8s-operators/infra-operator:latest
 IMG ?= $(DEFAULT_IMG)
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
-ENVTEST_K8S_VERSION = 1.25.0
+ENVTEST_K8S_VERSION = 1.26
```
stuggi, line 53: should be 1.26.1?

kajinamit: No this is intentional.

setup-envtest first attempts to find the exact version then it looks for the latest version matching the pattern. So This is now resolved to 1.26.1. I case a new bugfix version is released then we can use it without updating the version again, and I think that would be a better option for us.

See https://github.com/openstack-k8s-operators/keystone-operator/pull/293 .

stuggi: > No this is intentional.
> 
> setup-envtest first attempts to find the exact version then it looks for the latest version matching the pattern. So This is now resolved to 1.26.1. I case a new bugfix version is released then we can use it without updating the version again, and I think that would be a better option for us.
> 
> See [openstack-k8s-operators/keystone-operator#293](https://github.com/openstack-k8s-operators/keystone-operator/pull/293) .

Cool, thanks for explanation, this makes sense. then we should change it in keystone-operator https://github.com/openstack-k8s-operators/keystone-operator/pull/292/files to 1.26?

stuggi (approved):
/lgtm

Makefile
kajinamit: Yes and that's done by the 293 PR (which you already approved).

stuggi: right, sorry about that. should have read more carefully

kajinamit: NP . Thanks for reviews !

PR #107, opened 2023-08-07
Add ClusterIPs to dnsmasq service status

rabi:
This is required when some pods want to use the internal dns server and they can't reach the metallb endpoint with the nads on those networks.


controllers/network/dnsmasq_controller.go
```diff
@@ -445,6 +446,14 @@ func (r *DNSMasqReconciler) reconcileNormal(ctx context.Context, instance *netwo
 	//Update status with LoadBalancerIPs
 	instance.Status.DNSAddresses = instance.Spec.ExternalEndpoints[0].LoadBalancerIPs
 
+	//Update status with Cluster Addresses
+	poolName := instance.Spec.ExternalEndpoints[0].IPAddressPool
+	serviceName := fmt.Sprintf("%s-%s-%s", dnsmasq.ServiceName, instance.Name, poolName)
+	svc, err := service.GetServiceWithName(ctx, helper, serviceName, instance.Namespace)
+	if err == nil && svc != nil {
```
stuggi, line 453: should probably also handle `if err !=nil` ? otherwise it stops with reconcile successful.

rabi: ack, done

controllers/network/dnsmasq_controller.go
```diff
@@ -445,6 +446,20 @@ func (r *DNSMasqReconciler) reconcileNormal(ctx context.Context, instance *netwo
 	//Update status with LoadBalancerIPs
 	instance.Status.DNSAddresses = instance.Spec.ExternalEndpoints[0].LoadBalancerIPs
 
+	// Update status with Cluster Addresses
+	poolName := instance.Spec.ExternalEndpoints[0].IPAddressPool
+	serviceName := fmt.Sprintf("%s-%s-%s", dnsmasq.ServiceName, instance.Name, poolName)
+	svc, err := service.GetServiceWithName(ctx, helper, serviceName, instance.Namespace)
+
+	if err != nil {
+		util.LogErrorForObject(helper, err, fmt.Sprintf(
```
stuggi, line 455: that message is a duplicate. we already get that as the error message from https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/common/service/service.go#L261 . lets remove that and I think we are good to merge it

stuggi (approved):
/lgtm

stuggi:
baremetal provisioning needs this change, merging this as the job will fail

PR #108, opened 2023-08-08
Stop registering schemas for default objects

kajinamit:
Currently we register schemas for corev1 and appsv1 but this is incomplete and we need a few more schemas like batchv1. However we have not noticed it because the resources available in k8s by default don't need explicit schema registration.

Let's remove these instead of keeping the incomplete list.

### DIFF
--- a/tests/functional/suite_test.go
+++ b/tests/functional/suite_test.go
@@ -35,10 +35,6 @@ import (
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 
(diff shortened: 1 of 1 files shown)


kajinamit:
This change does not affect zuul jobs and we should skip these. I've pushed https://github.com/openstack-k8s-operators/infra-operator/pull/109 .

abays (approved):
/lgtm

abays:
recheck

kajinamit:
@abays I think we can override the job results. The change in envtest does not affect these deployment jobs. These jobs would be skipped automatically once https://github.com/openstack-k8s-operators/infra-operator/pull/109 is merged.

PR #109, opened 2023-08-08
Stop executing zuul jobs in test code change

kajinamit:
We don't have to run zuul jobs when the change affects only kuttl tests or env tests.


abays (approved):
/lgtm

kajinamit:
recheck

let me re-run the job once... but the change does not affect the functionality so we can override the result if some of these zuul jobs fail again.

PR #110, opened 2023-08-17
Fix link and maturity in csv

kajinamit:
The existing link is invalid. Also maturity was not consistent with API version(v1beta) currently presented by this operator.


abays (approved):
/lgtm

PR #114, opened 2023-08-28
Allow no-negcache option in dnsmasq

cjeanner:
This option allows to not cache the NXDOMAIN nor any other negative answer from the forwarders.

This is especially interesting when a DNS query happens when an operator endpoint isn't 100% ready, meaning its DNS record isn't known from the internal openshift-dns services.

Without that option, the dnsmasq service will cache the NXDOMAIN for the default TTL duration, meaning it will serve the cached answer instead of checking back.

### DIFF
--- a/controllers/network/dnsmasq_controller.go
+++ b/controllers/network/dnsmasq_controller.go
@@ -475,9 +475,12 @@ func (r *DNSMasqReconciler) generateServiceConfigMaps(
 
 	var cfg string
 	for _, option := range instance.Spec.Options {
-		cfg += option.Key + "="
-		values := strings.Join(option.Values, ",")
-		cfg += values + "\n"
+		cfg += option.Key
+		if len(option.Values) > 0 {
+			values := strings.Join(option.Values, ",")
+			cfg += "=" + values
+		}
+		cfg += "\n"
 	}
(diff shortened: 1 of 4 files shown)


pablintino:
recheck

abays (approved):
/lgtm

controllers/network/dnsmasq_controller.go
```diff
@@ -475,9 +475,12 @@ func (r *DNSMasqReconciler) generateServiceConfigMaps(
...
 	for _, option := range instance.Spec.Options {
-		cfg += option.Key + "="
-		values := strings.Join(option.Values, ",")
-		cfg += values + "\n"
+		cfg += option.Key
+		if len(option.Values) > 0 {
+			values := strings.Join(option.Values, ",")
+			cfg += "=" + values
+		}
+		cfg += "\n"
```
gibizer, line 483: You could add a test case in https://github.com/openstack-k8s-operators/infra-operator/blob/main/tests/functional/dnsmasq_controller_test.go covering both case: i) option with values, ii) option without values

cjeanner: Ah, was wondering if there were any tests... lemme see if I'm able to produce one. My golang is... well. what it is. Level-0

PR #115, opened 2023-08-29
Add missing tests for no-negcache option

cjeanner:
Following #114, we need to ensure the formatting is correct for option without any value such as the new no-negcache one we added.

### DIFF
--- a/tests/functional/base_test.go
+++ b/tests/functional/base_test.go
@@ -70,6 +70,10 @@ func GetDefaultDNSMasqSpec() map[string]interface{} {
 			Key:    "server",
 			Values: []string{"1.1.1.1"},
 		},
+		{
+			Key:    "no-negcache",
+			Values: []string{},
+		},
 	})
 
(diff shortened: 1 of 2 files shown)


tests/functional/base_test.go
```diff
@@ -70,6 +70,9 @@ func GetDefaultDNSMasqSpec() map[string]interface{} {
 			Key:    "server",
 			Values: []string{"1.1.1.1"},
 		},
+    {
```
gibizer, line 73: Indentation feels off at least on the github interface.

gibizer (commented):
Thanks @cjeanner! Content looks good but I think there is an indentation issue.

tests/functional/base_test.go
cjeanner: not only that, but I missed the closing bracket...

gibizer (approved):
Thanks! Looks good!

abays (approved):
/lgtm

PR #120, opened 2023-09-12
Switch infra-operator to use the multinode layout

pablintino:
Since we are getting rid off the nested EDPM jobs we are now changing the infra-operator to use the multinode based EDPM deployment job.
At the same time, we are applying the EDPM jobs using a job template that allows centrilized manangment of the jobs definition and improves maintainability.


raukadah (approved):

abays (approved):
/lgtm

abays:
recheck

PR #132, opened 2023-10-11
[envtest]Use lib-common/module/common/test/helper

gibizer:
The lib-common/module/test/helper is removed. It was an empty wrapper around lib-common/module/common/test/helper. So this patch switch to use that.


go.mod
```diff
@@ -22,7 +21,6 @@ require (
 
 require (
 	github.com/beorn7/perks v1.0.1 // indirect
-	github.com/cert-manager/cert-manager v1.11.5 // indirect
```
stuggi, line 25: I think we can now also remove the replace at the end of the file
```
// needed to to cert-manager v1.11.4 see https://github.com/cert-manager/cert-manager/blob/v1.11.4/go.mod#L263C1-L264C104
// remove this once we bump to cert-manager v1.12.x
replace github.com/Venafi/vcert/v4 => github.com/jetstack/vcert/v4 v4.9.6-0.20230519122548-219f317ae107 //allow-merging
```

gibizer: good point.

stuggi (approved):
/lgtm

PR #135, opened 2023-10-18
Allow setting local domains

rabi:
Allow setting local domains and not always forward queries to upstream servers.


stuggi (approved):
/lgtm

PR #141, opened 2023-10-29
Refactor use of pointer library

bshephar:
Ths pointer library[1] has been deprecated and replaced by ptr. This change refactors the usage on the new library[2] and usage.

[1] https://pkg.go.dev/k8s.io/utils/pointer
[2] https://pkg.go.dev/k8s.io/utils/ptr


abays (approved):
/lgtm

PR #149, opened 2023-11-17
Fix bundle validations (add sample reservation)

dprince:
This PR fixes some bundle validations:

WARN[0000] Skipping definitions parsing for API {}: Go type not found
WARN[0000] Warning: Value network.openstack.org/v1beta1, Kind=Reservation: provided API should have an example annotation

Adds a sample for the Reservation CR. Even though this is only used internally it won't hurt to add it.

Additionally, removes the 'service' definition from PROJECT as it only has a controller.


stuggi (approved):
/lgtm

PR #153, opened 2023-12-04
[transporturl] support TLS, watch name rabbitmq cluster resource and EnvTests

stuggi:
**support tls query**
The rabbitmq transport url requested needs to be in the following format for tls:
`rabbit://<user>:<pwd>@<service>:5671/?ssl=1`

- pass query `?ssl=1/0` depending of enabled disabled
- When tls is enabled a different port is used by the cluster. The information is part of the rabbitmq secret. Lets get the port information always from the secret.

**watch for named rabbit cluster**
When updates are performed to the rabbitmq cluster, like change from non tls to tls configuration the transporturl has to reconcile. This adds functionality to the transporturl controller to watch for the named rabbitmq cluster.

**Add envtest for default and tls use case**

Jira: [OSPRH-1919](https://issues.redhat.com//browse/OSPRH-1919)


stuggi:
recheck

stuggi:
recheck

stuggi:
recheck

PR #174, opened 2024-02-01
Add TLS support for Redis

dciabrin:
Ability to configure Redis with TLS endpoints and TLS replication for Redis HA.

Jira: [OSPRH-3567](https://issues.redhat.com//browse/OSPRH-3567)


pkg/redis/volumes.go
```diff
@@ -102,10 +121,45 @@ func getVolumes(r *redisv1beta1.Redis) []corev1.Volume {
...
+		}
+	}
+
+	return vols
+}
+
+func getTLSVolumeMounts(r *redisv1beta1.Redis) []corev1.VolumeMount {
+	var vols []corev1.VolumeMount
+	if r.Spec.TLS.Enabled() {
+		svc := tls.Service{
```
olliewalsh, line 146: why not use `r.Spec.TLS.GenericService.ToService()`?

pkg/redis/volumes.go
```diff
@@ -102,10 +121,45 @@ func getVolumes(r *redisv1beta1.Redis) []corev1.Volume {
...
+		if r.Spec.TLS.Ca.CaBundleSecretName != "" {
+			caVolume := r.Spec.TLS.Ca.CreateVolume()
+			vols = append(vols, caVolume)
+		}
+	}
+
+	return vols
+}
+
+func getTLSVolumeMounts(r *redisv1beta1.Redis) []corev1.VolumeMount {
```
stuggi, line 143: in the controller you changed it to `redisv1`, should we do this everywhere?

controllers/redis/redis_controller.go
```diff
@@ -180,9 +201,37 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ct
 		return rbacResult, nil
 	}
 
+	// Hash of all resources that may cause a service restart
+	inputHashEnv := make(map[string]env.Setter)
+
+	// Check and hash inputs
+	var certHash, caHash string
+	specTLS := &instance.Spec.TLS
+	if err == nil && specTLS.Enabled() {
```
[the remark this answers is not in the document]
dciabrin: You're correct, that's a leftover from a previous development. Thanks

pkg/redis/volumes.go
dciabrin: Right, I'll replicate the change everywhere, I think this is the pattern that other operator are using.

stuggi: yes, then you do not have to update it everywhere when there is an api version bump

stuggi (approved):
/lgtm

PR #175, opened 2024-02-01
Add TLS support for memcached

dciabrin:
Ability to configure memcached with TLS endpoints.

Jira: [OSPRH-3568](https://issues.redhat.com//browse/OSPRH-3568)


controllers/memcached/memcached_controller.go
```diff
@@ -185,9 +207,37 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ct
 		return rbacResult, nil
 	}
 
+	// Hash of all resources that may cause a service restart
+	inputHashEnv := make(map[string]env.Setter)
+
+	// Check and hash inputs
+	var certHash, caHash string
+	specTLS := &instance.Spec.TLS
+	if err == nil && specTLS.Enabled() {
```
stuggi, line 216: same as for redis, I think the `err == nil` can be removed

controllers/memcached/memcached_controller.go
```diff
@@ -185,9 +207,37 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ct
...
+		certHash, _, err = specTLS.GenericService.ValidateCertSecret(ctx, helper, instance.Namespace)
+		inputHashEnv["Cert"] = env.SetValue(certHash)
+	}
+	if err == nil && specTLS.Ca.CaBundleSecretName != "" {
+		caName := types.NamespacedName{
+			Name:      specTLS.Ca.CaBundleSecretName,
+			Namespace: instance.Namespace,
+		}
+		caHash, _, err = tls.ValidateCACertSecret(ctx, helper.GetClient(), caName)
```
stuggi, line 225: just realized this with how errors are handled here. the check for `err != nil` happens bellow and we still update the hash when the error happens in next line. in case of an existing deployment where everything is ok, if there is an error returned, e.g. a temporary issue with reading from etcd, we can see the hash flipping as it will be temporary at least to be set to "". since you use `certHash` and `caHash` vars we could just update the `inputHashEnv` after the `err != nil` check.

stuggi: same applies to redis https://github.com/openstack-k8s-operators/infra-operator/blob/main/controllers/redis/redis_controller.go#L220

controllers/memcached/memcached_controller.go
```diff
@@ -199,6 +249,27 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ct
...
+	if hashMap, changed := util.SetHash(instance.Status.Hash, common.InputHashName, hashOfHashes); changed {
+		// Hash changed and instance status should be updated (which will be done by main defer func),
+		// so update all the input hashes and return to reconcile again
+		instance.Status.Hash = hashMap
+		for k, s := range inputHashEnv {
+			var envVar corev1.EnvVar
+			s(&envVar)
+			instance.Status.Hash[k] = envVar.Value
+		}
+		util.LogForObject(helper, fmt.Sprintf("Input hash changed %s", hashOfHashes), instance)
```
stuggi, line 269: we should stop using `util.LogForObject` and instead use the `Log.Info()`. same for redis https://github.com/openstack-k8s-operators/infra-operator/blob/main/controllers/redis/redis_controller.go#L263

controllers/memcached/memcached_controller.go
dciabrin: Right I overlook the fact that the defer function will kick in even when we return an error. Should be fixed now.

controllers/memcached/memcached_controller.go
dciabrin: Ack, will do the other controllers as well (redis and galera)

dciabrin:
recheck

stuggi:
recheck

stuggi:
recheck

stuggi:
recheck

stuggi (approved):
/lgtm

PR #178, opened 2024-02-13
[DNSMasq]Check delete before creating ServiceAccount

gibizer:
When the namespace is deleted ServiceAccount creation will fail and because that runs before checking if the DNSMasq is being deleted too the DNSMasq will never be deleted. This patch moves the check before the SA creation.

Fixes: [OSP-27793](https://issues.redhat.com//browse/OSP-27793)


gibizer:
Together with https://github.com/openstack-k8s-operators/keystone-operator/pull/370 this fixes the namespace hanging issue

abays (approved):
/lgtm

stuggi (approved):
/lgtm

PR #191, opened 2024-02-28
Add *Core structs for redis,netconfig,memcached

dprince:
Adds a version of the structs (called "core") is meant to be used via the OpenStackControlplane. It is the same as the normal Spec's for redis, netconfig, and memcached only it is missing the containerImages.

The Default() function for webhooks has been updated accordingly.

Jira: [OSPRH-4835](https://issues.redhat.com//browse/OSPRH-4835)


stuggi (approved):
/lgtm

PR #192, opened 2024-03-06
[tlse] add TLS status config, func to return server list and tls support

stuggi:
Adds a TLSSupport to the memcached status which reflects if the instance got configured with TLS. Same flag was also introduced to mariadbdatabase a while back.
Also
* adds functions to return the memcached server list and the status of tls support, which can be used for clients for easy consumption. They could be used as an intermediate step for https://github.com/openstack-k8s-operators/infra-operator/pull/190
* creates the service list for memcached fqdn svc hostnames to be able to use fqdn hostnames in tls certs.

Jira: [OSPRH-5283](https://issues.redhat.com//browse/OSPRH-5283)

### DIFF
--- a/apis/bases/memcached.openstack.org_memcacheds.yaml
+++ b/apis/bases/memcached.openstack.org_memcacheds.yaml
@@ -134,6 +134,9 @@ spec:
                 items:
                   type: string
                 type: array
+              tlsSupport:
+                description: Whether TLS is supported by the memcached instance
+                type: boolean
             type: object
         type: object
     served: true
--- a/apis/memcached/v1beta1/memcached_funcs.go
+++ b/apis/memcached/v1beta1/memcached_funcs.go
@@ -1,0 +1,95 @@
+/*
+Copyright 2023.
+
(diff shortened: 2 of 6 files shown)


stuggi:
recheck

stuggi:
recheck

stuggi:
recheck

apis/bases/memcached.openstack.org_memcacheds.yaml
```diff
@@ -134,6 +134,9 @@ spec:
                 items:
                   type: string
                 type: array
+              tlsSupport:
+                description: Whether TLS is supported by the memcached instance
+                type: boolean
```
olliewalsh, line 139: maybe an enum to indicate the verify mode?

stuggi: as discussed on slack, if we introduce client certs, we could change it. its a status and atm not consumed direct. instead its consumed by keystone using the func added in this PR

PR #193, opened 2024-03-07
Replacing optional pointer to bool with optional bool

jpodivin:
Pointer doesn't really have any tangible benefits in this case. Rather, it forces us to implement checks where we wouldn't have to otherwise.


jpodivin:
/retest

timeout

apis/network/v1beta1/ipset_types.go
```diff
@@ -38,7 +38,7 @@ type IPSetNetwork struct {
 
 	// +kubebuilder:validation:Optional
 	// Use gateway from subnet as default route. There can only be one default route defined per IPSet.
-	DefaultRoute *bool `json:"defaultRoute,omitempty"`
+	DefaultRoute bool `json:"defaultRoute,omitempty"`
```
stuggi, line 41: if we want to do this change I think we should remove `omitEmpty` and probably explicit default it to false, which is basically the bool default. 

@rabi @hjensas you have thoughts on doing this change?

hjensas: I agree that we should remove `omitEmpty` and default it to `false`.

Another question is, do we need it? I know I asked for it during review, but looking at the actual implementations in dataplane-operator / edpm-ansible I don't see this hint used. 

In TrieplO we used it in nic-config templates and in THT to build the `_host_routes` routes variables.

For example in THT overcloud.j2.yaml 
```
          {%- if role.default_route_networks is not defined or 'ControlPlane' in role.default_route_networks %}
            - - default: true
                next_hop: {get_attr: [Networks, net_attributes_map, ctlplane, subnets, {get_param: {{role.name}}ControlPlaneSubnet}, gateway_ip]}
          {%- endif %}
```

I would have assumed this be used in `dataplane-operator/pkg/deployment/inventory.go#L186` or something - but it is not.

Should we remove `DefaultRoute` bool instead?

stuggi: it is used in the ipset https://github.com/openstack-k8s-operators/infra-operator/blob/main/controllers/network/ipset_controller.go#L446 to set the gateway information on the reservation. The inventory is then using this info https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/pkg/deployment/inventory.go#L203

rabi: >if we want to do this change I think we should remove omitEmpty and probably explicit default it to false, which is basically the bool default.

I thought we used bool pointers at many places to allow us distinguish between a boolean field explicitly set to false and one that has not been set at all.  boolean fields (when not using pointers) always default to false, so probably don't have to do explicit `default=false`. Yeah `omitempty` has to be dropped, though I'm not sure if we really need this change.

jpodivin: Omitempty is generally useless except when it comes to struct pointers. It can be said to have some, minimal, benefit when it comes to json handling but that's it. Since everybody agrees I'm dropping it entirely.

stuggi: > > if we want to do this change I think we should remove omitEmpty and probably explicit default it to false, which is basically the bool default.
> 
> I thought we used bool pointers at many places to allow us distinguish between a boolean field explicitly set to false and one that has not been set at all. boolean fields (when not using pointers) always default to false, so probably don't have to do explicit `default=false`. Yeah `omitempty` has to be dropped, though I'm not sure if we really need this change.

yes this is correct. in this case I think we could have it either true/false. I think the not set at all case is, at least right now, not required. But I agree that this change is not needed.

jpodivin:
Job failed with failed to call webhook baremetalhost.metal3.io error. Investigating

apis/network/v1beta1/ipset_types.go
stuggi: > Omitempty is generally useless except when it comes to struct pointers. It can be said to have some, minimal, benefit when it comes to json handling but that's it. Since everybody agrees I'm dropping it entirely.

this is not true. it has value for ptr types to not be included if not wanted. As Rabi said this change is not really needed. I'd just keep what we have. there is not real benefit from changing it.

jpodivin: I suppose if we really want to turn binary (true, false) into ternary (true, false, nil), then it makes sense to have boolean pointers as omit empty. I don't see much reason to keep that sort of complicated setup. But I will defer to you.

jpodivin:
There is a body of opinion against API simplification.

PR #195, opened 2024-03-07
Simplification of webhook conditionals

jpodivin:
Some of the conditionals were needlessly checking for nil value in cases already covered by previous checks.


jpodivin:
recheck

The deployment fails due to undefined var in the ansible role. There have been some changes recently in that area, so I will give this one more try. ` "'neutron_physical_bridge_name' is undefined. 'neutron_physical_bridge_name' is undefined"`

stuggi (approved):
/lgtm

PR #198, opened 2024-03-12
[memcached] add funcs to return quoted serverList and update test helpers

stuggi:
For tls, in https://github.com/openstack-k8s-operators/infra-operator/pull/192 changed the server lists to contain the svc fqdn, but missed to update the test helpers.


abays (approved):
/lgtm

abays:
recheck

Issue #200, opened 2024-03-12
infra-operator-controller-manager CrashLoopBackOff panic: runtime error: invalid memory address or nil pointer dereference


Loumy:
in openstack-operator the infra-operator-controller-manager crash when baremetal deploy start

```
2024-03-12T18:22:03Z    INFO    Observed a panic in reconciler: runtime error: invalid memory address or nil pointer dereference        {"controller": "ipset", "controllerGroup": "network.openstack.org", "controllerKind": "IPSet", "IPSet": {"name":"edpm-compute-01","namespace":"openstack"}, "namespace": "openstack", "name": "edpm-compute-01", "reconcileID": "5d768fc0-07ab-445a-ad16-22597016b542"}
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
        panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x17d32a9]

goroutine 331 [running]:
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile.func1()
        /opt/app-root/src/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.5/pkg/internal/controller/controller.go:116 +0x1fa
panic({0x19ccdc0, 0x2b8ea00})
        /usr/lib/golang/src/runtime/panic.go:884 +0x213
github.com/openstack-k8s-operators/infra-operator/controllers/network.(*IPSetReconciler).ensureReservation(0xc00030dd10, {0x1eb4410, 0xc002a30240}, 0xc002a0c2c0, 0xc0028ff7c0, 0xc00050c2c0, 0xc000479ab0)
        /remote-source/controllers/network/ipset_controller.go:441 +0x949
github.com/openstack-k8s-operators/infra-operator/controllers/network.(*IPSetReconciler).reconcileNormal(0xc00030dd10, {0x1eb4410, 0xc002a30240}, 0xc002a0c2c0, 0xc002a0c2c0?)
        /remote-source/controllers/network/ipset_controller.go:254 +0x2a9
github.com/openstack-k8s-operators/infra-operator/controllers/network.(*IPSetReconciler).Reconcile(0xc00030dd10, {0x1eb4410, 0xc002a30240}, {{{0xc000c108f6?, 0x5?}, {0xc000c10900?, 0xc000e5fd48?}}})
[...]

PR #201, opened 2024-03-13
Ensure we don't append to nil routes

rabi:
Also, handle ipv6 scenario.


rabi:
recheck RETRY_LIMIT

rabi:
Failed to run 'make crc_storage' target. Aborting

stuggi (approved):
/lgtm 

Thanks Rabi, this should prevent the panic in https://github.com/openstack-k8s-operators/infra-operator/issues/200

PR #202, opened 2024-03-13
[ipset] add validation in ipset webhook for default gw to be configured

stuggi:
This allows to only request the default gw for a network if it is configured on the referenced subnet.

Closes: https://github.com/openstack-k8s-operators/infra-operator/issues/200


stuggi:
recheck

abays (approved):
/lgtm

PR #205, opened 2024-03-27
Pin setup-envtest to last working commit

jpodivin:
Change[0] in controller-runtime is causing our tests to fail with go version mismatch.

This version restriction should be removed once the issue is fixed in the controller-runtime

[0] https://github.com/kubernetes-sigs/controller-runtime/pull/2693


abays (approved):
/lgtm

PR #212, opened 2024-04-11
Allow memcached to listen to both TLS and non-TLS

dciabrin:
When memcached is configured for exposing its endpoint over TLS, some clients still require accessing in via non-TLS. Update the TLS configuration to expose the TLS endpoint on port 11212 while keeping non-TLS accesses on port 11211. This matches what was configured by TripleO.

Updated kuttl tests to avoid multiple TestAssert in the same test as only the last one is executed. Also rename the assert files to give time to the resource to become ready before running the assert scripts to avoid spurious logs.

Jira: [OSPRH-3568](https://issues.redhat.com//browse/OSPRH-3568)


stuggi:
recheck

stuggi:
keystone bootstrap jobs fail. we try to get the service operator bump in the openstack-operator landed to have a possible fix for mariadb added.

stuggi:
recheck

dciabrin:
New revision to fix infra-operator-build-deploy-kuttl failure, as spotted by @stuggi

stuggi (approved):
/lgtm

PR #217, opened 2024-04-25
TransportURL: rely on PatchInstance instead of manual Update

gibizer:
The transporturl  controller did a separate client.Update after
it initialized it status then continued the normal reconciliation. As it
is explicitly updated the status subresource and no finalizer is added
this probably works. However to allow easier reasoning about when the
instance is updated this patch removes the explicit Update call and
instead returns immediately to use the deferred PatchInstance call
as the only place that persists the instance.


gibizer:
I mixed things up. I need to rethink the reasoning to remove the extra update, but the issue is not as sever as I thought here.

stuggi (approved):
/lgtm

gibizer:
recheck
the zuul trigger is missed due to a zuul restart

gibizer:
recheck retry limit

PR #221, opened 2024-04-30
ensure network order is preserved

SeanMooney:
This change ensure that the ipset network order
is preserved in the status.reservation

This is requried to ensure the dns search domain order
is defined by the order of the network list.

Closes: [OSPRH-6672](https://issues.redhat.com//browse/OSPRH-6672)

### DIFF
--- a/controllers/network/ipset_controller.go
+++ b/controllers/network/ipset_controller.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"fmt"
 	"net"
-	"sort"
 
 	corev1 "k8s.io/api/core/v1"
 	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
@@ -274,11 +273,6 @@ func (r *IPSetReconciler) reconcileNormal(ctx context.Context, instance *network
 
 			return ctrl.Result{}, err
 		}
-
-		// sort instance.Status.Reservations by Network
-		sort.Slice(instance.Status.Reservation, func(i, j int) bool {
-			return instance.Status.Reservation[i].Network < instance.Status.Reservation[j].Network
-		})
 
 		instance.Status.Conditions.MarkTrue(networkv1.ReservationReadyCondition, networkv1.ReservationReadyMessage)
 
--- a/tests/functional/ipset_controller_test.go
+++ b/tests/functional/ipset_controller_test.go
@@ -17,6 +17,8 @@ limitations under the License.
 package functional_test
 
 import (
+	"math/rand"
+
 	. "github.com/onsi/ginkgo/v2" //revive:disable:dot-imports
 	. "github.com/onsi/gomega"    //revive:disable:dot-imports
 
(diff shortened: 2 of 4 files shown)


tests/functional/ipset_controller_test.go
```diff
@@ -312,6 +314,72 @@ var _ = Describe("IPSet controller", func() {
...
 	})
 
+	When("an IPSet with multiple networks gets created", func() {
+		var netSpecs []networkv1.Network
+		var networks []networkv1.IPSetNetwork
+		BeforeEach(func() {
+			netSpecs = []networkv1.Network{}
+			netSpecs = append(netSpecs, GetNetSpec(net1, GetSubnet1(subnet1)))
+			netSpecs = append(netSpecs, GetNetSpec(net2, GetSubnet2(subnet1)))
+			netSpecs = append(netSpecs, GetNetSpec(net3, GetSubnet3(subnet1)))
```
SeanMooney, line 324: i am not sure if the netspecs are required by the way.

i assume  yes as all the other test create both

i assume i need to create the network specification to then be able to assign IP from those network to the ipsets.

so i think this is required but just said i would call this out incase i can simplfy

controllers/network/ipset_controller.go
```diff
@@ -275,11 +274,6 @@ func (r *IPSetReconciler) reconcileNormal(ctx context.Context, instance *network
 			return ctrl.Result{}, err
 		}
 
-		// sort instance.Status.Reservations by Network
-		sort.Slice(instance.Status.Reservation, func(i, j int) bool {
-			return instance.Status.Reservation[i].Network < instance.Status.Reservation[j].Network
-		})
-
```
SeanMooney, line 282: this is the actuall fix

tests/functional/ipset_controller_test.go
```diff
@@ -312,6 +314,72 @@ var _ = Describe("IPSet controller", func() {
...
+		})
+
+		It("it should have created an IPSet with multiple networks", func() {
+
+			instance := &networkv1.IPSet{}
+
+			Eventually(func(g Gomega) {
+				g.Expect(k8sClient.Get(ctx, ipSetName, instance)).Should(Succeed())
+				for i := 0; i < len(networks); i++ {
+					g.Expect(instance.Spec.Networks[i].Name).To(Equal(networks[i].Name))
```
stuggi, line 374: this validates that the instance.spec.networks[X] is the same as the defined networks[X].

tests/functional/ipset_controller_test.go
```diff
@@ -312,6 +314,72 @@ var _ = Describe("IPSet controller", func() {
...
+			netSpecs = []networkv1.Network{}
+			netSpecs = append(netSpecs, GetNetSpec(net1, GetSubnet1(subnet1)))
+			netSpecs = append(netSpecs, GetNetSpec(net2, GetSubnet2(subnet1)))
+			netSpecs = append(netSpecs, GetNetSpec(net3, GetSubnet3(subnet1)))
+			networks = []networkv1.IPSetNetwork{}
+			networks = append(networks, GetIPSetNet1Lower())
+			networks = append(networks, GetIPSetNet2())
+			networks = append(networks, GetIPSetNet3())
+
+			Expect(networks).To(HaveLen(len(netSpecs)))
```
stuggi, line 330: don't think this is a must. like you could have more networks in the netSpecs (osnetconfig) then in the ipset networks you request IPs for.

tests/functional/ipset_controller_test.go
```diff
@@ -312,6 +314,72 @@ var _ = Describe("IPSet controller", func() {
...
+			// we want to ensure order is preserved so we will randomize the order of the networks
+			// and then create the IPSet
+			rand.Shuffle(len(networks), func(i, j int) {
+				networks[i], networks[j] = networks[j], networks[i]
+				netSpecs[i], netSpecs[j] = netSpecs[j], netSpecs[i]
+			})
+			// confirm that after we randomize the order the networks and netSpec are still in sync
+			for i := 0; i < len(networks); i++ {
+				Expect(networks[i].Name).To(Equal(netSpecs[i].Name))
+			}
```
stuggi, line 340: same this must not be the case in a deployment. one could have 3 networks in netSpecs (netconfig), but only 2 of them in the networks of the ipset.

tests/functional/ipset_controller_test.go
```diff
@@ -312,6 +314,72 @@ var _ = Describe("IPSet controller", func() {
...
+
+		It("it should have created an IPSet with multiple networks", func() {
+
+			instance := &networkv1.IPSet{}
+
+			Eventually(func(g Gomega) {
+				g.Expect(k8sClient.Get(ctx, ipSetName, instance)).Should(Succeed())
+				for i := 0; i < len(networks); i++ {
+					g.Expect(instance.Spec.Networks[i].Name).To(Equal(networks[i].Name))
+					g.Expect(instance.Spec.Networks[i].Name).To(Equal(netSpecs[i].Name))
```
stuggi, line 375: Don't think this is valid/needed. it works for this test implementation, but could be different in a real env. This validates that the network referenced in instance.Spec.Networks[i] is at the same place as in the osnetconfig. Lets assume you create the netconfig with 3 networks, but the ipset just requests 2 of them, so it could be index 1 in Networks, but index 2 in the osnetconfig.

tests/functional/ipset_controller_test.go
```diff
@@ -312,6 +314,72 @@ var _ = Describe("IPSet controller", func() {
...
+					g.Expect(instance.Spec.Networks[i].Name).To(Equal(instance.Status.Reservation[i].Network))
```
stuggi, line 376: +2 this is the important one

controllers/network/ipset_controller.go
stuggi: +2 as we discussed on slack

stuggi:
adding a hold to land it after the beta freeze

stuggi:
> added some comments and removed the asserts that were not needed otheresie the core is the same. ill check back in a few days if there are any other comments and respin but since this will not be in beta there is no rush.

thanks @SeanMooney looks good to me. lets get that in after the freeze

stuggi (approved):
/lgtm

Issue #233, opened 2024-06-25
DualStack Openshift causes the memcached listen IPs to be wrongly formatted


blackphantom39:
Hello,

I'm currently testing the 18.0-Beta of the RHOSO deployment and ran into the following issue creating the control plane and therefore the memcached pods.

Given the PodIPs look like this:

```yaml
  podIPs:
    - ip: 10.130.1.23
    - ip: 'fd74:ae5d:e44d:1::1826'
```

the resulting environment variable for POD_IPS will look like: `POD_IPS=10.130.0.198,fd74:ae5d:e44d:1::19b5` this will not work correctly when inserted in the [template](https://github.com/openstack-k8s-operators/infra-operator/blob/main/templates/memcached/config/memcached) for the LISTEN_IPS. The resulting LISTEN variable will have a malformed IPv6 causing the pod to crash.

Resulting LISTEN:

```
echo $LISTEN
127.0.0.1 notls:127.0.0.1:11211 [::1] notls:[::1]:11211 [10.130.0.198,fd74:ae5d:e44d:1::19b5] notls:[10.130.0.198,fd74:ae5d:e44d:1::19b5]:11211
                                                        └─────── not correctly split ───────┘
```

blackphantom39:
It should be sufficient to just replace the reference to $POD_IPS by either of the following:

`$(printf %s $POD_IPS | tr ',' ' ')` resulting in:
```
LISTEN=$(echo 127.0.0.1 ::1 $(printf %s $POD_IPS | tr ',' ' ') | tr ' ' '\n' | sed 's/\(.*\):\(.*\)/[\1:\2]/' {{ .memcachedTLSListen }})
```

or  `$(echo $POD_IPS | tr ',' ' ')` resulting in:
```
LISTEN=$(echo 127.0.0.1 ::1 $(echo $POD_IPS | tr ',' ' ') | tr ' ' '\n' | sed 's/\(.*\):\(.*\)/[\1:\2]/' {{ .memcachedTLSListen }})
```

stuggi:
@blackphantom39 thanks for reporting this bug. you want to submit a PR to fix this?

stuggi:
just fyi, there will probably be follow up issues in other components as for GA dual stack was not targeted as a supported feature.

blackphantom39:
Closed as of PR #236

PR #236, opened 2024-06-25
memcached: replace ',' in $POD_IPS by space

blackphantom39:
Adjust the handling of the `$POD_IPS` variable to allow for multiple pod IPs. This change takes the comma separated list of PodIPs and replaces the comma with a space to make it work with the rest of the statement for building the listen IPs.

Implements #233

### DIFF
--- a/templates/memcached/config/memcached
+++ b/templates/memcached/config/memcached
@@ -3,5 +3,5 @@ USER="memcached"
 MAXCONN="8192"
 CACHESIZE="9932"
 # explicit IP to bind to (wrap IPv6 in brackets)
-LISTEN=$(echo 127.0.0.1 ::1 $(printf %s $POD_IPS | tr ',' ' ') | tr ' ' '\n' | sed 's/\(.*\):\(.*\)/[\1:\2]/' {{ .memcachedTLSListen }})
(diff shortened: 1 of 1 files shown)


templates/memcached/config/memcached
```diff
@@ -3,5 +3,5 @@ USER="memcached"
 MAXCONN="8192"
 CACHESIZE="9932"
 # explicit IP to bind to (wrap IPv6 in brackets)
-LISTEN=$(echo 127.0.0.1 ::1 $POD_IPS | tr ' ' '\n' | sed 's/\(.*\):\(.*\)/[\1:\2]/' {{ .memcachedTLSListen }})
+LISTEN=$(echo 127.0.0.1 ::1 $(printf %s $POD_IPS | tr ',' ' ') | tr ' ' '\n' | sed 's/\(.*\):\(.*\)/[\1:\2]/' {{ .memcachedTLSListen }})
```
dciabrin, line 6: The printf here is probably not necessary, as we only have the localhost addresses before $POD_IPS and we don't plan on adding more hard coded values here.
What about just keeping too tr sequentially instead? It should be sufficient
```
echo 127.0.0.1 ::1 $POD_IPS | tr ',' ' ' | tr ' ' '\n'
```

blackphantom39: works for me. will submit the change

PR #245, opened 2024-07-11
[memcached] add CR name length validation

stuggi:
The memcached controller creates StatefulSet for the memcached to run. This adds a StatefulSet pod's label
"controller-revision-hash": "<statefulset_name>-<hash>" to the pod. The kubernetes label is restricted under 63 char and the revision hash is an int32, 10 chars + the hyphen. With this the max name must not exceed 52 chars.

Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/532

Jira: https://issues.redhat.com/browse/OSPRH-8063


tests/functional/memcached_webhook_test.go
```diff
@@ -0,0 +1,65 @@
...
+					"name":      "foo-1234567890-1234567890-1234567890-1234567890-1234567890",
+					"namespace": namespace,
+				},
+				"spec": GetDefaultMemcachedSpec(),
+			}
+
+			unstructuredObj := &unstructured.Unstructured{Object: raw}
+			_, err := controllerutil.CreateOrPatch(
+				th.Ctx, th.K8sClient, unstructuredObj, func() error { return nil })
+			Expect(err).To(HaveOccurred())
```
gibizer, line 62: if you want to assert the message then you can use the following pattern to extract it from the error:
https://github.com/openstack-k8s-operators/nova-operator/blob/232a1f1b9ef305d2718ad728720edae03b4322d6/test/functional/validation_webhook_test.go#L125-L137

gibizer (approved):
looks good to me. Thanks!

PR #271, opened 2024-09-11
Add ServiceNetwork attribute to networks in netconfig

rabi:
This would allow us to identify the ctlplane network when
using custom network and also generate inventory using
the ServiceNetwork.

Jira: https://issues.redhat.com/browse/OSPRH-10125

### DIFF
--- a/apis/bases/network.openstack.org_netconfigs.yaml
+++ b/apis/bases/network.openstack.org_netconfigs.yaml
@@ -55,6 +55,10 @@ spec:
                         ...
                       pattern: ^[a-zA-Z0-9][a-zA-Z0-9\-_]*[a-zA-Z0-9]$
                       type: string
+                    serviceNetwork:
+                      description: Service network mapping
+                      pattern: ^[a-z0-9][a-z0-9\-_]*[a-z0-9]$
+                      type: string
                     subnets:
                       description: Subnets of the network
                       items:
--- a/controllers/network/ipset_controller.go
+++ b/controllers/network/ipset_controller.go
@@ -20,6 +20,8 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"sort"
+	"strings"
 
 	corev1 "k8s.io/api/core/v1"
 	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
@@ -42,6 +44,8 @@ import (
 	helper "github.com/openstack-k8s-operators/lib-common/modules/common/helper"
 	util "github.com/openstack-k8s-operators/lib-common/modules/common/util"
 )
+
+const CtlPlaneNetwork = "ctlplane"
 
 // IPSetReconciler reconciles a IPSet object
 type IPSetReconciler struct {
@@ -274,6 +278,13 @@ func (r *IPSetReconciler) reconcileNormal(ctx context.Context, instance *network
 			return ctrl.Result{}, err
 		}
 
+		// sort instance.Status.Reservations
+		sort.Slice(instance.Status.Reservation, func(i, j int) bool {
+			return (strings.EqualFold(string(instance.Status.Reservation[i].ServiceNetwork),
+				CtlPlaneNetwork) && !strings.EqualFold(string(instance.Status.Reservation[j].ServiceNetwork),
+				CtlPlaneNetwork))
+		})
+
 		instance.Status.Conditions.MarkTrue(networkv1.ReservationReadyCondition, networkv1.ReservationReadyMessage)
 
 		Log.Info("IPSet is ready:", "instance", instance.Name, "ipSetRes", ipSetRes.Spec.Reservation)
@@ -398,6 +409,10 @@ func (r *IPSetReconciler) ensureReservation(
 			return nil, err
 		}
 
+		if netDef.ServiceNetwork == "" {
+			netDef.ServiceNetwork = networkv1.ToDefaultServiceNetwork(netDef.Name)
+		}
+
 		// set net: subnet label
 		reservationLabels = util.MergeStringMaps(reservationLabels,
 			map[string]string{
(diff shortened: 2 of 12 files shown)


rabi:
We're discussing whether or how to support custom networks.

stuggi:
as discussed, I think an alternative approach would be to do https://github.com/openstack-k8s-operators/infra-operator/pull/272 and limit the network names. the subnet names can still be customized.

controllers/network/ipset_controller.go
```diff
@@ -274,6 +278,13 @@ func (r *IPSetReconciler) reconcileNormal(ctx context.Context, instance *network
 			return ctrl.Result{}, err
 		}
 
+		// sort instance.Status.Reservations by Network
+		sort.Slice(instance.Status.Reservation, func(i, j int) bool {
+			return (strings.EqualFold(instance.Status.Reservation[i].ServiceNetwork,
```
stuggi, line 283: in existing env there is no ServiceNetwork. we probably have to fallback in the ipset controller to the network name if thats the case

rabi:
recheck

rabi:
recheck `Somthing is not deployed in CRC. Exit!` error message is not very helpful :smile:

config/manager/kustomization.yaml
```diff
@@ -4,5 +4,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: quay.io/openstack-k8s-operators/infra-operator
+  newName: quay.io/rabi/infra-operator
```
bshephar, line 7: Needs to be reverted

controllers/network/ipset_controller.go
```diff
@@ -43,6 +45,8 @@ import (
 	util "github.com/openstack-k8s-operators/lib-common/modules/common/util"
 )
 
+const CtlPlaneNetwork = "ctlplane"
```
bshephar, line 48: nit: I don't think this needs to be Heap allocated right? Looks like we're only using this const in the `reconcileNormal` function, so the const can be stack allocated in that function.

Unless you were planning to use it externally somewhere.

bshephar (dismissed):
Just the quay.io reference that actually needs to be changed. My other comment isn't overly important, but I left it there for consideration.

controllers/network/ipset_controller.go
rabi: `const` is a compile time construct in go, so there is no question heap/stack memory allocation for them at run time. They're logically replaced in code at compile time.

config/manager/kustomization.yaml
rabi: Why is it even included? not in `.gitignore` I guess.

controllers/network/ipset_controller.go
rabi: ack, done!

tests/functional/ipset_controller_test.go
```diff
@@ -369,11 +369,22 @@ var _ = Describe("IPSet controller", func() {
...
 					g.Expect(instance.Spec.Networks[i].Name).To(Equal(ipSetNetworks[i].Name))
 					// then assert that the reservation networks are in the same order
-					g.Expect(instance.Spec.Networks[i].Name).To(Equal(instance.Status.Reservation[i].Network))
+					// other than ctlplane network being the first one.
+					if ipSetNetworks[i].Name == ctlplaneNetwork {
+						ctlplaneFound = true
+						continue
+					}
+					if ctlplaneFound {
+						g.Expect(instance.Spec.Networks[i].Name).To(Equal(instance.Status.Reservation[i].Network))
```
stuggi, line 384: maybe we verify that `i`  must be `0` for the ctlplane case? `g.Expect(i).To(Equal(0))`

apis/bases/network.openstack.org_netconfigs.yaml
```diff
@@ -55,6 +55,10 @@ spec:
                         ...
                       pattern: ^[a-zA-Z0-9][a-zA-Z0-9\-_]*[a-zA-Z0-9]$
                       type: string
+                    serviceNetwork:
+                      description: Service network mapping
+                      pattern: ^[a-zA-Z0-9][a-zA-Z0-9\-_]*[a-zA-Z0-9]$
```
stuggi, line 60: should the servicenetwork forced to be all lower case, and document the internal defaults (ctlplane, internalapi, external, storage, storagemgmt and tenant)?

tests/functional/ipset_controller_test.go
rabi: The test does that to identify the ctlplane network in L372, right?

apis/bases/network.openstack.org_netconfigs.yaml
rabi: Network and hence SeviceNetwork are case insensitive as before. Ctlplane and ctlplane mean the same. It's handled in code.

rabi: What benefits do you think we would get with that? We already allow case insensitivity of `Network` after https://github.com/openstack-k8s-operators/infra-operator/commit/0cab900cc4025768b5b48fda396efb1ee5f92fd1 and never made it case-sensitive. We can allow them to be used as ServiceNetwork without case conversion in defaulting. Also, we  use mixed case in many places like https://github.com/openstack-k8s-operators/infra-operator/blob/main/config/samples/network_v1beta1_ipset.yaml#L7

tests/functional/ipset_controller_test.go
stuggi: right, with L372 and the condition in L379 it would be the same what I commented, but I don't think we have a test that the network[0] ServiceNetwork is "ctlplane", or do we? maybe we should have a validation that `instance.Status.Reservation[0].ServiceNetwork == "ctlplane"`?

apis/bases/network.openstack.org_netconfigs.yaml
stuggi: I was just wondering if it makes sense to force the user to use lowercase for the service network, otherwise the user might wonder why different service network name would be used when have to debug. right now you'd have to know that it will be made lowercase by the code. that was just what I was thinking.

stuggi: ah wait, the servicenetwork in the reservation will be automatic lower case. I think we are good. sorry

stuggi (approved):
/lgtm

PR #272, opened 2024-09-11
[netconfig] limit network names to be one of the known

stuggi:
network names should be one of
- ctlplane
- internalapi
- external
- tenant
- storage
- storagemgmt

The dataplane is creating the inventory based on the network name and passes ip information as ansible vars like e.g. tenant_ip.

If a custom network name is used, as a follow up custom information needs to be passed into the services. Therefore only allow network names to be one of the expected, while the subnet names can be customized.

### DIFF
--- a/apis/bases/network.openstack.org_netconfigs.yaml
+++ b/apis/bases/network.openstack.org_netconfigs.yaml
@@ -51,8 +51,15 @@ spec:
                       description: MTU of the network
                       type: integer
                     name:
-                      description: Name of the network, e.g. External, InternalApi,
-                        ...
+                      description: Name of the network, could be one of ctlplane,
+                        internalapi, external, storage, storagemgmt or tenant
+                      enum:
+                      - ctlplane
+                      - internalapi
+                      - external
+                      - storage
+                      - storagemgmt
+                      - tenant
                       pattern: ^[a-zA-Z0-9][a-zA-Z0-9\-_]*[a-zA-Z0-9]$
                       type: string
                     subnets:
--- a/apis/network/v1beta1/netconfig_types.go
+++ b/apis/network/v1beta1/netconfig_types.go
@@ -31,7 +31,8 @@ type NetNameStr string
 // Network definition
 type Network struct {
 	// +kubebuilder:validation:Required
-	// Name of the network, e.g. External, InternalApi, ...
+	// +kubebuilder:validation:Enum=ctlplane;internalapi;external;storage;storagemgmt;tenant
+	// Name of the network, could be one of ctlplane, internalapi, external, storage, storagemgmt or tenant
 	Name NetNameStr `json:"name"`
 
 	// +kubebuilder:validation:Required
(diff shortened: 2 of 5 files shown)


stuggi:
if custom network names are used you'd have to check any role and pass in the correct information like e.g. https://github.com/openstack-k8s-operators/edpm-ansible/blob/c21563b42bcb15c6ee1c16cc991a9b2717e24728/roles/edpm_ovn/defaults/main.yml#L54

rabi:
> if custom network names are used you'd have to check any role and pass in the correct information like e.g. https://github.com/openstack-k8s-operators/edpm-ansible/blob/c21563b42bcb15c6ee1c16cc991a9b2717e24728/roles/edpm_ovn/defaults/main.yml#L54

As discussed in slack, we don't have to do above, if  we've some kind of mapping in netconfig and use it. Though I'm ok, if we decide not to support custom network names and it has no future implications. 

apiVersion: network.openstack.org/v1beta1
kind: NetConfig
metadata:
  name: netconfig
spec:
  networks:
  - name: net1
    serviceNet: ctlplane
    ...
 - name: net2
   serviceNet: internalapi

stuggi:
ok this version will introduce an issue with https://github.com/openstack-k8s-operators/architecture/blob/a2f05988589a26c4a12d660c236b8cf337c1683b/examples/va/hci/edpm-pre-ceph/nodeset/values.yaml#L125-L126 where a custom swift network got added

then we could do either
- what Rabi proposed in https://github.com/openstack-k8s-operators/infra-operator/pull/271 with the downside that a user would have to know details of the roles and what vars of the role to updated with custom network name information.
- my other initial proposal when we were discussing this, to have a network type parameter which you can set on the network to reflect that this is ctlplane, tenant, internalapi or whatever. if no type is set the network name is used in the reservation.

What I do not like is that a user would have to read all the roles to see what information should be  passed in when we do not have the default network names.

@abays @rabi @hjensas what are your thoughts?

stuggi:
> > if custom network names are used you'd have to check any role and pass in the correct information like e.g. https://github.com/openstack-k8s-operators/edpm-ansible/blob/c21563b42bcb15c6ee1c16cc991a9b2717e24728/roles/edpm_ovn/defaults/main.yml#L54
> 
> As discussed in slack, we don't have to do above, if we've some kind of mapping in netconfig and use it. Though I'm ok, if we decide not to support custom network names and it has no future implications.
> 
> apiVersion: network.openstack.org/v1beta1 kind: NetConfig metadata: name: netconfig spec: networks:
> 
> * name: net1
>   serviceNet: ctlplane
>   ...
> * name: net2
>   serviceNet: internalapi

yes thats basically same as the network type I mentioned initially

hjensas:
So we don't have composable networks and a service net map to allow the user to specify what network to use for what service like we did with tripleo? If we lock down the names like this how do we add for example a network used for Manila NFS (What was called `GaneshaNetwork` in tripleo)?

stuggi:
> So we don't have composable networks and a service net map to allow the user to specify what network to use for what service like we did with tripleo? If we lock down the names like this how do we add for example a network used for Manila NFS (What was called `GaneshaNetwork` in tripleo)?

yes thats a use case I was not thinking about. @rabi has a new proposal in https://github.com/openstack-k8s-operators/infra-operator/pull/271 to add a ServiceNetwork parameter.

PR #273, opened 2024-09-12
Rename InstanceHA to InstanceHa and use configmap for containerImage

lmiccini:
This commit renames the InstanceHA occurrences into InstanceHa to comply with the openstack-operator's openstackversion naming (case) convention.

We also take the chance and use a configmap created by the openstack-operator to get the containerImage, so to orchestrate its update.


lmiccini:
recheck

stuggi (approved):
/lgtm

PR #281, opened 2024-10-01
Wrap reverse dns lookup in try/except block

lmiccini:
If DNS is broken calling socket.gethostbyaddr() could result in an unexpected pod crash.
Let's wrap the call in try/except with the expectation that a reverse dns lookup failure should not prevent evacuation, worst case scenario we would not capture the memory dump and the user will have a "ERROR Could not perform reverse dns lookup for: X" in the logs hinting at resolution not working.

### DIFF
--- a/templates/instanceha/bin/instanceha.py
+++ b/templates/instanceha/bin/instanceha.py
@@ -419,7 +419,11 @@ def _check_kdump(stale_services):
         #logging.debug("address is %s" % address[0])
 
         # short hostname
-        name = socket.gethostbyaddr(address[0])[0].split('.', 1)[0]
+        try:
+            name = socket.gethostbyaddr(address[0])[0].split('.', 1)[0]
+        except Exception as msg:
+            logging.error('Failed reverse dns lookup for: %s - %s' % (address[0], msg))
+            continue
 
         # fence_kdump checks if the magic number matches, so let's do it here too
         if hex(struct.unpack('ii',data)[0]).upper() != FENCE_KDUMP_MAGIC.upper() :


templates/instanceha/bin/instanceha.py
```diff
@@ -419,7 +419,11 @@ def _check_kdump(stale_services):
         #logging.debug("address is %s" % address[0])
 
         # short hostname
-        name = socket.gethostbyaddr(address[0])[0].split('.', 1)[0]
+        try:
+            name = socket.gethostbyaddr(address[0])[0].split('.', 1)[0]
+        except Exception as msg:
```
jpodivin, line 424: It would be better to catch something more specific than general exception.

jpodivin (commented):
As long as the `t_end` isn't too far in the future this should work fine.
Otherwise I would add retry limit or something. 

lgtm, but someone with more familiary should check too

templates/instanceha/bin/instanceha.py
lmiccini: With the generic exception we will catch:
ERROR:root:Failed reverse dns lookup for: 8.8.8.88 - [Errno 1] Unknown host (valid dns server, wrong or unknown ip)
ERROR:root:Failed reverse dns lookup for: 8.8.8.88 - [Errno 2] Host name lookup failure (dns server unreachable or bogus response)
that is the same returned by except socket.herror or similar afaik.

lmiccini:
obsolete, will be fixed in a single PR tackling two more issues we found.

PR #285, opened 2024-10-11
Remove KeystoneApi dependency and improve InstanceHa Conditions messages

lmiccini:
This commit improves the conditions messages so to make the status more explicit, for example returning a status like:

[zuul@controller-0 ~]$ oc get instanceha
NAME           STATUS   MESSAGE
instanceha-0   False    Instance HA waiting for Fencing secret

when the Fencing Secret has not been provided or created yet.


stuggi (approved):
/lgtm

PR #288, opened 2024-10-15
Increase POLL interval to avoid race with check_kdump

lmiccini:
This commit increases the default value of 'POLL' from 30s to 45s. 
The reason is that the check_kdump functionality has an internal hard-coded timeout of 30s and this may induce race conditions if POLL is set to the same value.
A warning has been added to make sure users are notified if POLL=30s.

Finally, we take the chance to normalize the retcode for bmh_fencing.


stuggi (approved):
/lgtm could also check the config via the iha controller and set a warning or info condition?

PR #289, opened 2024-10-16
Automate OLM cleanup for running local operator w/ webhooks

abays:
When using make `run-with-webhook`, local versions of the operator and its webhooks are added to the cluster. If the operator was previously installed via OLM, then there might be lingering webhooks from that installation. We've previously been assuming that the user would manually remove them, but we should just get rid of them ourselves since those OLM webhooks need to be deleted anyhow for the local webhooks to function unimpeded.  We can also automatically scale down the operator's OLM deployment in such a scenario.

Also adds missing local webhooks for `InstanceHA` resources.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -359,16 +359,11 @@ operator-lint: gowork ## Runs operator-lint
 	go vet -vettool=$(LOCALBIN)/operator-lint ./... ./apis/...
 
 # Used for webhook testing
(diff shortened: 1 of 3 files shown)


stuggi (approved):
/lgtm

lmiccini (approved):
thanks Andrew!

PR #295, opened 2024-10-22
Add delay parameter

lmiccini:
We've seen issues in the past with nfs servers that would not release locks fast enough for evacuation to complete (see https://bugzilla.redhat.com/show_bug.cgi?id=1755760 for example). 
This commit adds a DELAY parameter so we can wait for a certain amount of time before starting the evacuation.


stuggi (approved):
/lgtm

PR #296, opened 2024-10-23
Fix typo / make bundle issue

lmiccini:
Spotted a place where I missed a 'HA' vs "Ha', result is that make bundle would generate a duplicate definition in config/manifests/bases/infra-operator.clusterserviceversion.yaml without this patch.


abays (approved):
/lgtm

PR #322, opened 2024-12-02
[bgp] Add new CRD to manage FRRConfiguration

stuggi:
Secondary network interfaces on pods need be announced in a BGP environment. This can be done by creating FRRConfiguration, per default in the metallb namespace.

This PR introduce a new CRD which, if an instance got created, the controller watches pods
* which have the NAD annotation on it
* the NAD has an IPAM configured

For each of them a FRRConfiguration gets created. The metallbs k8s service FRRConfiguration of that worker node is taken as the base to create this configuration.

Known issue:
Since there are then two FRRConfiguration, which hold same configs, like timeouts. It is not possible to update thise. The FRRConfiguration webhook will block those. A possible way to change it, would be stop the infra-operator controller-manager, delete the pod FRRConfigurations, do the change that it gets reflected in the metallb LB FRRConfiguration, then enable the controller that the pod FRRConfiguration get re-created.

Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/588
Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/589

Jira: [OSPRH-8680](https://issues.redhat.com//browse/OSPRH-8680)
Jira: [OSPRH-12384](https://issues.redhat.com/browse/OSPRH-12384)


stuggi:
recheck

lmiccini (approved):

PR #332, opened 2025-01-09
Make cacheSize maxConn configurable

xek:
Jira: [OSPRH-12339](https://issues.redhat.com//browse/OSPRH-12339)

### DIFF
--- a/apis/bases/memcached.openstack.org_memcacheds.yaml
+++ b/apis/bases/memcached.openstack.org_memcacheds.yaml
@@ -48,10 +48,20 @@ spec:
           spec:
             description: MemcachedSpec defines the desired state of Memcached
             properties:
+              cacheSize:
+                default: 9932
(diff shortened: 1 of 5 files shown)


config/samples/memcached_v1beta1_memcached.yaml
```diff
@@ -4,3 +4,4 @@ metadata:
   name: memcached
 spec:
   replicas: 1
+  cacheSize: 1024
```
dciabrin, line 7: Is that required? I haven't tried the patch, but I would expect no change in memory behavior without it.

xek: I'll remove it, I think I just used this for testing.

dciabrin (approved):

abays (approved):
/lgtm

PR #334, opened 2025-01-13
DNS: Support large DNS payloads

Akrog:
When edpm nodes use the dnsmasq we deploy in OCP they will not be able to get large DNS responses.

DNS messages are restricted to 512 octets, and tools such as dig will try detect that the response has been truncated and then use TCP mode instead.

We can see this happen if we run inside the edpm node the following:

```
$ dig cisco.com TXT
;; Truncated, retrying in TCP mode.
```

With this patch we make the dns OCP service expose the TCP port so the alternative TCP mechanism can be used.


csibbitt (approved):
LGTM - hoping there isn't some larger story to why this is currently missing

stuggi (approved):
/lgtm thanks Gorka!

stuggi:
the failing test is not related. I'll have a look into that. I think we have seen this in the past, looks like an issue in that bgpconfiguration functional test.

stuggi:
GET https://quay.io/v2/: unexpected status code 502 Bad Gateway

PR #335, opened 2025-01-17
Drop the default containerImage parameter and use VAR

lmiccini:
Hard coding the containerImage value is problematic b/c our build system would not override the value set in the crd automatically, so the variable ends up being set to the wrong image in the crd definition.
Let's try using RELATED_* vars (set by the infra-operator itself, because we can't have two vars using the same value, see https://github.com/openstack-k8s-operators/openstack-operator/pull/1150).

Jira: [OSPRH-13093](https://issues.redhat.com//browse/OSPRH-13093)

### DIFF
--- a/apis/bases/instanceha.openstack.org_instancehas.yaml
+++ b/apis/bases/instanceha.openstack.org_instancehas.yaml
@@ -53,7 +53,6 @@ spec:
                   bundle file
                 type: string
               containerImage:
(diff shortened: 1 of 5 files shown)


bshephar (approved):

lmiccini:
recheck

stuggi (approved):
/lgtm

PR #339, opened 2025-01-24
Update disable_reason when evacuation fails and other fixes

lmiccini:
This commit brings the following changes:

1. When evacuations fail we want to update the disable_log_reason field so users can immediately see the reason, plus we can filter out these computes and not try to evacuate them again.

2. We now filter out evacuations that are 'failed' when deciding if a compute can be re-enabled after it has been fenced, as it is preferable to avoid users the additional burden of manually re-enabling the hypervisors if not necessary.
If necessary the LEAVE_DISABLED option can be used to keep a host down.

3. We wrap the initial call to nova api in try/except to give users evidence in the logs if nova is not available.

4. We added a FORCE_ENABLE parameter. If true, ignore any failed evacuation and enable computes after evacuation.

Jira: https://issues.redhat.com/browse/OSPRH-11279


stuggi (approved):
/lgtm

PR #341, opened 2025-01-30
Allow enabling mTLS for memcached

lmiccini:
This commit reworks the TLS section in order to add two new parameters:
1. 'SslVerifyMode' allows users to enable/enforce mTLS for memcached
2. 'AuthCertSecret' sets the secret name containing the certificate
   that clients should use.
   This should be set as 'AuthCertSecret.SecretName'

See https://docs.memcached.org/features/tls/

By default SslVerfyMode is set to "None" or 0, but it can be set to "Request" or "Require".

If set to "Require" it will:
1. prevent the creation of the non-tls memcached svc port (11211)
2. enforce client side certification validation

Futher additions:
1. Status string 'MTLSCert' to store the secret name
2. Function GetMemcachedMTLSSecret() to give services a way to fetch and use the required certificate.


lmiccini:
recheck

stuggi:
recheck

controllers/memcached/memcached_controller.go
```diff
@@ -182,7 +180,12 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ct
 
 	instance.Status.Conditions.Init(&cl)
 	instance.Status.ObservedGeneration = instance.Generation
-	instance.Status.MTLSCert = ""
+
+	if instance.Spec.TLS.MTLS.SslVerifyMode == "Request" || instance.Spec.TLS.MTLS.SslVerifyMode == "Require" {
+		// MTLS cert secrets
+		cl = append(cl, *condition.UnknownCondition(memcachedv1.MTLSInputReadyCondition, condition.InitReason, condition.InputReadyInitMessage))
+		instance.Status.MTLSCert = ""
```
stuggi, line 187: now that I think of it again, I am not sure if we should reset it on each iteration. if we fail in the middle for some other reason, the name of the cert name of the previous run would be still valid, right?

controllers/memcached/memcached_controller.go
```diff
@@ -274,6 +280,42 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ct
...
+					condition.SeverityWarning,
+					condition.TLSInputErrorMessage,
+					err.Error()))
+				return ctrl.Result{}, err
+			}
+			inputHashEnv["ClientCert"] = env.SetValue(hash)
+			instance.Status.MTLSCert = *instance.Spec.TLS.MTLS.AuthCertSecret.SecretName
+			instance.Status.Conditions.MarkTrue(memcachedv1.MTLSInputReadyCondition, condition.InputReadyMessage)
+		} else {
+			instance.Status.Conditions.Set(condition.FalseCondition(
```
stuggi, line 309: we probably should set it to `instance.Status.MTLSCert = ""` in here?

controllers/memcached/memcached_controller.go
```diff
@@ -181,6 +181,12 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ct
 	instance.Status.Conditions.Init(&cl)
 	instance.Status.ObservedGeneration = instance.Generation
 
+	if instance.Spec.TLS.MTLS.SslVerifyMode == "Request" || instance.Spec.TLS.MTLS.SslVerifyMode == "Require" {
+		// MTLS cert secrets
+		cl = append(cl, *condition.UnknownCondition(memcachedv1.MTLSInputReadyCondition, condition.InitReason, condition.InputReadyInitMessage))
+		instance.Status.MTLSCert = ""
+	}
```
stuggi, line 188: we should set it to `instance.Status.MTLSCert = ""` in an else for switching to `None`

lmiccini: shall I extend this loop with the else or do we want to use the one validating the certificate?

stuggi: I think its ok to just add the else here

PR #342, opened 2025-02-10
[topology] move common funcs from lib-common to the operators

stuggi:
This removes a circular dependency between lib-common, common module, and the infra-operator.


fmount (approved):
/lgtm

olliewalsh (approved):
/lgtm

PR #345, opened 2025-02-13
Make InstanceHA ContainerImage optional

lmiccini:
With 502a989f7bbe8291e86750ce1d055feb43f22585 we removed the default value for ContainerImage but forgot to make the parameter optional. 
This commit rectifies this, so users can rely on the defaults set by the operator/webhook.


stuggi (approved):
/lgtm

PR #346, opened 2025-02-13
[topology] Add common logic to the inject TopologySpec

fmount:
Right now every operator that references a topology needs to have some common logic to process the retrieved CR Spec and inject to the resulting deployment.
It might be a good idea to hide this logic (and simplify the code maintenance) and move it to infra-operator API.
By doing this we can limit potential future updates to (all) services that consume a Topology CR by reference.

### DIFF
--- a/apis/topology/v1beta1/topology_types.go
+++ b/apis/topology/v1beta1/topology_types.go
@@ -211,3 +211,77 @@ func EnsureDeletedTopologyRef(
 	}
 	return ctrl.Result{}, nil
 }
+
+// Given a StatefulSet/Deployment/DaemonSet PodTemplateSpec, apply the referenced
+// topology
+func (t Topology) ApplyTo(
+	pod *corev1.PodTemplateSpec,
+) {
+	// Get the Topology .Spec
+	ts := t.Spec
+	// Process TopologySpreadConstraints if defined in the referenced Topology
+	if ts.TopologySpreadConstraints != nil {
+		pod.Spec.TopologySpreadConstraints = *t.Spec.TopologySpreadConstraints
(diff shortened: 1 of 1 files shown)


apis/topology/v1beta1/topology_types.go
```diff
@@ -211,3 +211,77 @@ func EnsureDeletedTopologyRef(
...
+	// Get the Topology .Spec
+	ts := t.Spec
+	// Process TopologySpreadConstraints if defined in the referenced Topology
+	if ts.TopologySpreadConstraints != nil {
+		pod.Spec.TopologySpreadConstraints = *t.Spec.TopologySpreadConstraints
+	}
+	// Process Affinity if defined in the referenced Topology
+	if ts.Affinity != nil {
+		pod.Spec.Affinity = ts.Affinity
+	}
```
fmount, line 229: @olliewalsh do you think we should include here the `DistributePods` part? I didn't add it because of DaemonSets that would use the same function, but let me know if you have thoughts around this

fmount:
recheck

stuggi:
recheck

gibizer (commented):
Looks good to me. Thanks!

stuggi:
looks good to me. lets get some feedback from @olliewalsh

apis/topology/v1beta1/topology_types.go
olliewalsh: No, I don't think we should include it, at least not in this function

olliewalsh (approved):
/lgtm

olliewalsh:
recheck

PR #350, opened 2025-02-20
Update LastAppliedTopology to TopologyRef interface

fmount:
This patch updates the SpecCore webhook interface to accept the Namespace as parameter. This change ensures that we properly validate the referenced topology from openstack-operator, rejecting requests with a different namespace. It also aligns "LastAppliedTopology" to "TopologyRef" interface. While TopologyRef is able to reference Name and namespace, we currently only save TopologyRef.Name to .Status.LastAppliedTopology. This patch aligns both interfaces to use the same type in the same form. In addition, some logic is moved to infra-operator because is common to all operators, and this patch also adopts this new form.

Depends-On: https://github.com/openstack-k8s-operators/openstack-operator/pull/1306


fmount:
recheck

fmount:
recheck

stuggi (approved):
/lgtm

PR #351, opened 2025-02-27
Create lightweight controller for rabbitmq service

olliewalsh:
For each rabbitmq instance we need a watch for it's topology reference.
We also need to track the last applied topology reference.

To support this, and to better align with all of the other services,
create a rabbitmq resource/controller that owns and manages the
underlying rabbitmq-cluster resource.

Related: [OSPRH-14301](https://issues.redhat.com//browse/OSPRH-14301)


controllers/rabbitmq/rabbitmq_controller.go
```diff
@@ -0,0 +1,494 @@
...
+
+	err = instance.Spec.MarshalInto(&rabbitmqCluster.Spec)
+	if err != nil {
+		instance.Status.Conditions.Set(condition.FalseCondition(
+			condition.ServiceConfigReadyCondition,
+			condition.ErrorReason,
+			condition.SeverityWarning,
+			condition.ServiceConfigReadyErrorMessage,
+			err.Error()))
+		return ctrl.Result{}, fmt.Errorf("error creating RabbitmqClusterS Spec: %w", err)
```
stuggi, line 293: nit RabbitmqClusterS

controllers/rabbitmq/rabbitmq_controller.go
```diff
@@ -0,0 +1,494 @@
...
+	topology, err := topologyv1.EnsureServiceTopology(
+		ctx,
+		helper,
+		instance.Spec.TopologyRef,
+		instance.GetLastAppliedTopologyRef(),
+		instance.Name,
+		metav1.LabelSelector{
+			// Default to the labels rabbitmqClusterOperator sets
+			MatchLabels: map[string]string{
+				"app.kubernetes.io/name":      instance.Name,
```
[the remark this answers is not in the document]
olliewalsh: taken from https://github.com/openstack-k8s-operators/rabbitmq-cluster-operator/blob/v2.9.0_patches/internal/metadata/label.go#L19
```
$ oc get statefulset rabbitmq-cell1-server --show-labels
NAME                    READY   AGE   LABELS
rabbitmq-cell1-server   1/1     10h   app.kubernetes.io/component=rabbitmq,app.kubernetes.io/name=rabbitmq-cell1,app.kubernetes.io/part-of=rabbitmq
```

controllers/rabbitmq/rabbitmq_controller.go
```diff
@@ -0,0 +1,494 @@
...
+	//
+	// Handle Topology
+	//
+	topology, err := topologyv1.EnsureServiceTopology(
+		ctx,
+		helper,
+		instance.Spec.TopologyRef,
+		instance.GetLastAppliedTopologyRef(),
+		instance.Name,
+		metav1.LabelSelector{
```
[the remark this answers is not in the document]
olliewalsh: Sure, will keep it consistent

controllers/rabbitmq/rabbitmq_controller.go
```diff
@@ -0,0 +1,494 @@
...
+	)
+
+	instance.Status.Conditions.Init(&cl)
+	instance.Status.ObservedGeneration = instance.Generation
+
+	// Init Topology condition if there's a reference
+	if instance.Spec.TopologyRef != nil {
+		c := condition.UnknownCondition(condition.TopologyReadyCondition, condition.InitReason, condition.TopologyReadyInitMessage)
+		cl.Set(c)
+	}
```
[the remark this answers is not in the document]
fmount: I think it's good to keep it here for now, and eventually review it and change for all the operators in a follow up.  In general you `InitConditions(` with the default (known) conditions, but Topology is set only if exists.
Here we're extracting from the InitCondition function the `cl.Set`  logic (that accepts the reference to the current condition) and it's the same thing that happens within `InitConditions`,
tl;dr we should be good here and we can follow up to make sure we have a consistent ordering.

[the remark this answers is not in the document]
stuggi: yes its not a big thing, but I think with this we never set the TopologyReadyCondition TopologyReadyInitMessage on the instance. we'd just set the the TopologyReadyCondition to whatever comes next in the flow

apis/rabbitmq/v1beta1/rabbitmq_webhook.go
```diff
@@ -0,0 +1,175 @@
...
+func (r *RabbitMq) ValidateUpdate(_ runtime.Object) (admission.Warnings, error) {
+	rabbitmqlog.Info("validate update", "name", r.Name)
+
+	var allErrs field.ErrorList
+	var allWarn []string
+	basePath := field.NewPath("spec")
+
+	// When a TopologyRef CR is referenced, fail if a different Namespace is
+	// referenced because is not supported
+	if r.Spec.TopologyRef != nil {
```
fmount, line 125: (for a follow up) Wondering if we should have a single SpecCore ValidateCreate and ValidateUpdate as part of the `RabbitMqSpec` that returns the SpecCore related function, so we only have to write it once. I saw this pattern in many operators, maybe we can think about it for a follow up.

apis/rabbitmq/v1beta1/webhook_suite_test.go
```diff
@@ -0,0 +1,145 @@
...
+
+var (
+	cfg       *rest.Config
+	k8sClient client.Client
+	testEnv   *envtest.Environment
+	ctx       context.Context
+	cancel    context.CancelFunc
+)
+
+func TestAPIs(t *testing.T) {
```
fmount, line 57: This is probably a placeholder, and the real testing happens under `tests/functional`. Do we want to have a webhook related test like we did for memcached ? [1]

[1] https://github.com/openstack-k8s-operators/infra-operator/blob/main/tests/functional/memcached_controller_test.go#L143

pkg/rabbitmq/cluster.go
```diff
@@ -0,0 +1,284 @@
...
+			cluster.Spec.Override.StatefulSet.Spec.Template.Spec.TopologySpreadConstraints = *topology.Spec.TopologySpreadConstraints
+		}
+		// Process Affinity if defined in the referenced Topology
+		if ts.Affinity != nil {
+			cluster.Spec.Override.StatefulSet.Spec.Template.Spec.Affinity = ts.Affinity
+		}
+	}
+	if cluster.Spec.Affinity == nil {
+		cluster.Spec.Affinity = &corev1.Affinity{
+			PodAntiAffinity: &corev1.PodAntiAffinity{
```
fmount, line 140: wondering if this is equals to [1], that we could use to remove some duplicated code

[1] https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/common/affinity/affinity.go#L26

fmount (commented):
I did a first pass. There are no blocking requests (especially due to the timeline that we have) and all my comments can be addressed in a potential follow up.

fmount:
we might need an update to the functional tests to run the new controller with the right dependency

pkg/rabbitmq/cluster.go
```diff
@@ -0,0 +1,284 @@
...
+
+	if cluster.Spec.Override.StatefulSet.Spec.Template.Spec.NodeSelector == nil {
+		if nodeselector != nil {
+			cluster.Spec.Override.StatefulSet.Spec.Template.Spec.NodeSelector = *nodeselector
+		}
+	}
+
+	if topology != nil {
+		// Get the Topology .Spec
+		ts := topology.Spec
```
fmount, line 128: should this be moved to `topology.ApplyTo(` pattern?

pkg/rabbitmq/cluster.go
```diff
@@ -0,0 +1,284 @@
...
+			inetProtocol,
+			tlsArgs,
+			fipsArgs,
+		),
+	}, corev1.EnvVar{
+		Name:  "RABBITMQ_CTL_ERL_ARGS",
+		Value: fmt.Sprintf("-proto_dist %s_%s %s", inetFamily, inetProtocol, tlsArgs),
+	})
+
+	defaultStatefulSet := rabbitmqv2.StatefulSet{
```
fmount, line 90: I'm not sure about that, but do we have a way to pass labels here?

pkg/rabbitmq/cluster.go
```diff
@@ -0,0 +1,284 @@
...
+	}
+	if cluster.Spec.Affinity == nil {
+		cluster.Spec.Affinity = &corev1.Affinity{
+			PodAntiAffinity: &corev1.PodAntiAffinity{
+				PreferredDuringSchedulingIgnoredDuringExecution: []corev1.WeightedPodAffinityTerm{
+					{
+						Weight: 100,
+						PodAffinityTerm: corev1.PodAffinityTerm{
+							LabelSelector: &metav1.LabelSelector{
+								MatchExpressions: []metav1.LabelSelectorRequirement{
```
fmount, line 146: if we create `serviceLabels` in the controller (that we use as `defaultLabelSelector` for topology), we could pass them down to the `StatefulSet` (but not sure it is possible for the `rabbitmqv2.StatefulSet`) we might have the same selectors here instead of `MatchExpression`, or do we have a different reason to only use `k8sAppName` ?

fmount: I think I answered my own question looking at https://github.com/rabbitmq/cluster-operator/blob/main/api/v1beta1/rabbitmqcluster_types.go and I see

```
  selector:
    matchLabels:
      app.kubernetes.io/name: rabbitmq
```

pkg/rabbitmq/impl/rabbitmqcluster.go
```diff
@@ -0,0 +1,119 @@
...
+	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+// NewRabbitMqCluster returns an initialized RabbitmqCluster.
+func NewRabbitMqCluster(
+	rabbit *rabbitmqv2.RabbitmqCluster,
```
fmount, line 20: @olliewalsh should we pass here the `serviceLabels` we generate from the controller and that are used as Topology defaultLabelSelector? I had this comment in a different (probably the wrong) place, but I think here is more appropriate.
I imagine the flow as:
1. we generate `serviceLabels` in rabbitmq_controller.go
2. we call `NewRabbitMqCluster` and we pass down those `serviceLlabels`
3. we assign them to `.Spec.Selector`.

what do you think about this? Does it make sense?

controllers/rabbitmq/rabbitmq_controller.go
```diff
@@ -0,0 +1,494 @@
...
+		// update the Status with the last retrieved TopologyRef
+		instance.Status.LastAppliedTopology = instance.Spec.TopologyRef
+		// update the TopologyRef associated condition
+		instance.Status.Conditions.MarkTrue(condition.TopologyReadyCondition, condition.TopologyReadyMessage)
+	} else {
+		// remove LastAppliedTopology from the .Status
+		instance.Status.LastAppliedTopology = nil
+	}
+
+	rabbitmqImplCluster := impl.NewRabbitMqCluster(rabbitmq.ConfigureCluster(rabbitmqCluster, IPv6Enabled, fipsEnabled, topology, instance.Spec.NodeSelector), 5)
```
fmount, line 339: We should pass to the `NewRabbitMqCluster` the `serviceLabels` used as `defaultLabelSelector` for the topology on L309, and update the `NewRabbitMqCluster` signature to accept them (and pass down to the `statefulSet .Spec.Selector`).

pkg/rabbitmq/cluster.go
olliewalsh: cannot use it, different type https://github.com/openstack-k8s-operators/rabbitmq-cluster-operator/blob/v2.9.0_patches/api/v1beta1/rabbitmqcluster_types.go#L380

pkg/rabbitmq/cluster.go
olliewalsh: Why? Probably can, need to look at the rabbitmqcluster logic

fmount: ignore this comment, sorry, this goes to the NewRabbitMqCluster function as selectors. I'm resolving now so it doesn't generate confusion

controllers/rabbitmq/rabbitmq_controller.go
olliewalsh: no, rabbitmqcluster sets the labels, do we just set defaultLabelSelector to the same values it uses

controllers/rabbitmq/rabbitmq_controller.go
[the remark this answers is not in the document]
stuggi: if you push another version you could just move this condition above L164, or we do  it in a follow up

controllers/rabbitmq/rabbitmq_controller.go
stuggi: we could use the override to change the default labels of the rabbimqcluster, BUT this will break existing deployments. so I think we can not do it.

pkg/rabbitmq/cluster.go
stuggi: yes I think we could use the func from lib-common. we'd just have to pass in the correct params.

controllers/rabbitmq/rabbitmq_controller.go
fmount: ok, so, no labels passing to the underlying statefulSet, +1 to not break any existing deployment. In this case, I think the `defaultLabelSelector` we pass to the Topology function (L308) should match the statefulset selector, otherwise it won't work: the `matchLabels` implements an "&&" condition and if the selector is different we have changes the topologySpreadConstraints is not properly applied during the scheduling.
In my environment I see:
```
}
{
  "matchLabels": {
    "app.kubernetes.io/name": "rabbitmq-cell1"
  }
}
{
  "matchLabels": {
    "app.kubernetes.io/name": "rabbitmq"
  }
}
```
so we should probably match only this one on L308, right?

olliewalsh: the docs here https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#components state the selector and pod template labels must be the same... but rabbitmqcluster doesn't do that

ultimately it doesn't really matter provided the selector includes the most specific label

olliewalsh:
```
$ oc get pods rabbitmq-cell1-server-0 --show-labels
NAME                      READY   STATUS    RESTARTS   AGE   LABELS
rabbitmq-cell1-server-0   1/1     Running   0          15h   app.kubernetes.io/component=rabbitmq,app.kubernetes.io/name=rabbitmq-cell1,app.kubernetes.io/part-of=rabbitmq,apps.kubernetes.io/pod-index=0,controller-revision-hash=rabbitmq-cell1-server-55fdbd897d,statefulset.kubernetes.io/pod-name=rabbitmq-cell1-server-0

```

pkg/rabbitmq/cluster.go
olliewalsh: sure, can address in a follow-up though unless I need to push another change here

olliewalsh:
> you cannot LGTM your own PR.

it's not going to do anything until the openstack-operator PR to use it land, so making an exception in this case :-)

PR #356, opened 2025-03-10
Add centralized topology validation function for service operators

fmount:
This patch introduces a unified topology validation function that service operators can use to validate `TopologyRef` definitions in their `CRs`.
 Key improvements:
- It creates a single entry point (that can be extended as needed) for all topology validation logic
- It accepts a `TopoRef` parameter and handles `nil` checks internally
- It validates `namespace` correctly without requiring `basePath` centralization (that lives in the service API definition)
- Fixes a bug in `ValidateTopologyNamespace` where a `topology` key was erroneously referenced (there's no topology entry in the CRD).

The new validation function provides a base that can be extended for more complex validation requirements in the future while maintaining consistent behavior across all service operators.

Jira: https://issues.redhat.com/browse/OSPRH-14626

### DIFF
--- a/apis/topology/v1beta1/topology_types.go
+++ b/apis/topology/v1beta1/topology_types.go
@@ -107,11 +107,24 @@ func GetTopologyByName(
 	return topology, hash, nil
 }
 
+// ValidateTopologyRef - returns a field.ErrorList when the Service references an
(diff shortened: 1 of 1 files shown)


stuggi (approved):
/lgtm

gibizer (commented):
Looks good to me too

PR #360, opened 2025-03-19
Fix topology namespace dereference

fmount:
This patch fixes how we dereference `topologyRef.namespace`. This is a struct and should have `.Child` to access the fields, and not `.Key` which suggests we're accessing an index of a map.

Jira: https://issues.redhat.com/browse/OSPRH-14626


gibizer (approved):
The fix is correct. In repositories where this is used and tested we expect that the tests needs to be adapted to this change along with the dependency bump. While this is an API change (the error message returned by the API is changing) it is a bugfix that worth risking a low chance breaking a tool depending on the old error message.

PR #367, opened 2025-04-02
Optimize EnsureDeletedTopologyRef condition

fmount:
As a follow up of [1] it is ok to simplify the service operators and blindly pass `instance.Status.LastAppliedTopology` to `EnsureServiceTopology` function provided by infra-operator.
The logic that removes the `finalizer` from an existing `TopologyRef` (only when `LastAppliedTopology != nil`) is now moved here, resulting in a simplification of the current operators logic.
This patch resolves yet another tech debt generated by the fact that we started with a different implementation of `LastAppliedTopology` (based on the `.Name` `string`), that later has been aligned with the existing `TopoRef` struct.

Jira: https://issues.redhat.com/browse/OSPRH-15072

[1] https://github.com/openstack-k8s-operators/placement-operator/pull/285#discussion_r1967176124

### DIFF
--- a/apis/topology/v1beta1/topology_types.go
+++ b/apis/topology/v1beta1/topology_types.go
@@ -266,8 +266,8 @@ func EnsureServiceTopology(
 	// 2. a topology reference is updated in the Service Component CR (tpRef != nil)
 	//    and the finalizer should be removed from the previously
 	//    referenced topology (tpRef.Name != lastAppliedTopology.Name)
-	if (tpRef == nil && lastAppliedTopology != nil) ||
-		(tpRef != nil && tpRef.Name != lastAppliedTopology.Name) {
+	if (lastAppliedTopology != nil) &&
+		(tpRef == nil || tpRef.Name != lastAppliedTopology.Name) {
 		_, err = EnsureDeletedTopologyRef(
 			ctx,
 			helper,
--- a/apis/memcached/v1beta1/memcached_types.go
+++ b/apis/memcached/v1beta1/memcached_types.go
@@ -159,19 +159,6 @@ func init() {
 	SchemeBuilder.Register(&Memcached{}, &MemcachedList{})
 }
 
-// GetLastAppliedTopologyRef - Returns the lastAppliedTopologyName that can be
-// processed by the handle topology logic
-func (instance Memcached) GetLastAppliedTopologyRef() *topologyv1.TopoRef {
-	lastAppliedTopologyName := ""
-	if instance.Status.LastAppliedTopology != nil {
-			lastAppliedTopologyName = instance.Status.LastAppliedTopology.Name
-	}
-	return &topologyv1.TopoRef{
-			Name:	   lastAppliedTopologyName,
-			Namespace: instance.Namespace,
-	}
-}
-
 // ValidateTopology -
 func (instance *MemcachedSpecCore) ValidateTopology(
 	basePath *field.Path,
--- a/apis/network/v1beta1/dnsmasq_types.go
+++ b/apis/network/v1beta1/dnsmasq_types.go
@@ -182,19 +182,6 @@ func SetupDefaults() {
 	SetupDNSMasqDefaults(dnsMasqDefaults)
 }
 
-// GetLastAppliedTopologyRef - Returns the lastAppliedTopologyName that can be
-// processed by the handle topology logic
-func (instance DNSMasq) GetLastAppliedTopologyRef() *topologyv1.TopoRef {
-	lastAppliedTopologyName := ""
-	if instance.Status.LastAppliedTopology != nil {
-			lastAppliedTopologyName = instance.Status.LastAppliedTopology.Name
-	}
-	return &topologyv1.TopoRef{
-			Name:	   lastAppliedTopologyName,
-			Namespace: instance.Namespace,
-	}
-}
-
 // ValidateTopology -
 func (instance *DNSMasqSpecCore) ValidateTopology(
 	basePath *field.Path,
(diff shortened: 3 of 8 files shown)


apis/topology/v1beta1/topology_types.go
```diff
@@ -266,8 +266,8 @@ func EnsureServiceTopology(
 	// 2. a topology reference is updated in the Service Component CR (tpRef != nil)
 	//    and the finalizer should be removed from the previously
 	//    referenced topology (tpRef.Name != lastAppliedTopology.Name)
-	if (tpRef == nil && lastAppliedTopology != nil) ||
-		(tpRef != nil && tpRef.Name != lastAppliedTopology.Name) {
+	if (lastAppliedTopology != nil) &&
```
fmount, line 269: - (**add** use case): Do not call `EnsureDeletedTopologyRef` (which removes the `finalizer`) is `lastAppliedTopology`, passed as input, is nil (e.g. we didn't have a referenced topology earlier, and this might be the first time we update the CR to add a new `topologyRef`)
- If there is a valid `lastAppliedTopology`, calling `EnsureDeletedTopologyRef` (which removes the `finalizer`) depends on the the fact that it doesn't match with `topologyRef.Name` (the current referenced topology). If it doesn't match, we have to remove the `finalizer` from `lastAppliedTopology` in two cases:
  - (**delete** use case): `topologyRef` is `nil` (the topology has been removed from the CR, hence we need to remove the finalizer)
  - (**update** use case): `topologyRef.Name` is different from `lastAppliedTopology.Name` (the topology has been updated, and the reference has been changed to point to a different topology instance)

fmount:
FYI I still need to make some testing to make sure everything still works as expected.

fmount:
> FYI I still need to make some testing to make sure everything still works as expected.

Done

stuggi (commented):
looks good to me

apis/topology/v1beta1/topology_types.go
gibizer: Thanks for the explanation it really helps understanding the condition. I tried to come up with a bit more readable code that requires less explanation by adding additional symbols and moving conditions out of the way if possible. What do you think about the following structure?

```golang
// No refs nothing to change or return
if (tpRef == nil && lastAppliedTpRef == nil){
   return nil, nil
}

// default input namespace
// ? can we do this in the defaulting webhook instead?
if (tpRef != nil && tpRef.Namespace == "" ) {
    tpRef.Namespace = helper...GetNamespace()
}

refChanged = ( !equal(tpRef, lastAppliedTpRef))

if (refChanged && lastAppliedTpRef != nil) {
    EnsureDeletedTopologyRef(lastAppliedTpRef)
}

if (tpRef != nil) {
    EnsureToplogyRef(tpRef)
}

```
Where `equal` can rely on structural equality of values via:
```golang
func equal(a *Ref, b *Ref) bool {
	if a == nil && b == nil {
		return true
	}

	if a == nil || b == nil {
		return false
	}

	return *a == *b // go will do structural equality of struct values
}
```
See it working in https://go.dev/play/p/LcGxSrOxsA7

fmount:
ok @gibizer @stuggi after landing this I will update nova and glance that I already proposed, and I'm then going to create the remaining changes for all the other operators.

apis/topology/v1beta1/topology_types.go
fmount: Hey @gibizer, I think yes it's more readable and probably makes sense. My goal here is to minimize the amount of changes and rely to what we tested as much as possible. While you did a great exercise of reworking the requirements into a simple and more readable algorithm, I'm not sure I have right now enough time to go through a full retest (because I usually build the operators on top of this change, rebuild openstack-operator and put everything together).
I'm ok to put this on hold and work more on it if you think we might hit bugs and reworking it has more benefits, otherwise I would go with the current (with more comments based on my previous comment) version.

gibizer: I'm OK to keep it as is if others are OK too. 

> I'm not sure I have right now enough time to go through a full retest

This signals to me that we either not have enough automated integration tests in place to avoid manual testing needs or we are not trusting that automated test. In both case I suggest to put extra effort into improving the integration tests of this feature so that we can avoid the very costly manual testing in the future.

gibizer (approved):

apis/topology/v1beta1/topology_types.go
fmount: > I'm OK to keep it as is if others are OK too.
> 
> > I'm not sure I have right now enough time to go through a full retest
> 
> This signals to me that we either not have enough automated integration tests in place to avoid manual testing needs or we are not trusting that automated test. In both case I suggest to put extra effort into improving the integration tests of this feature so that we can avoid the very costly manual testing in the future.

I agree with you and the cost of a manual testing is not entirely insignificant.
There's a plan to have automated testing via DT [1][2]. With an available DT I could simply feed it with my own `-index`(by passing an override to ci-framework) and do a sanity check over pending patches based on the some content deployed by a custom `openstack-operator` bundle.
Unfortunately this is planned for FR3 (along with other work) and we're currently not that far. I preferred to proactively start filling this tech debt instead of potentially waiting months and leave this work not uncompleted.
I'll see what I can do but for now I would go w/ the amount of changes that minimize the amount of time spent on tech debt, until we have more people or more automation in place :D

[1] https://github.com/fultonj/architecture/tree/dz
[2] https://github.com/fultonj/architecture/commit/6285b9c2e621e0993ce15afaf649ebf2c5053c42

PR #369, opened 2025-04-11
Add TAGGED_AGGREGATES parameter and cleanups

lmiccini:
This commit:

1 Adds a new parameter called TAGGED_AGGREGATES (false by default): this allows operators to enable instanceha only on aggregates marked with the EVACUABLE_TAG metadata, and also to disable/enable this functionality without having to update aggregates.

2 Cleans up a few pieces of code and adds a more explicit and informative warning message if a non-tagged compute has failed.

3 Improves how the reserved_hosts are used. Now we try to enable a compute from the same aggregate of the failed compute (if TAGGED_AGGREGATES are used) or a compute from the same AZ.

4 Fixes a potential issue if fencing is not working, preventing evacuation.


stuggi (approved):
/lgtm

PR #380, opened 2025-04-23
Enforce ha-all policy to make queues mirrored

lmiccini:
This commits adds a new parameter QueueType (default=Mirrored) controlling the eventual enforcement of a policy like:

```
vhost   name    pattern apply-to        definition      priority
/       ha-all          all     {"ha-mode":"exactly","ha-params":2,"ha-promote-on-shutdown":"always"}   0
```

This is to temporarily (sic) help services to survive a rabbitmq disruption or rolling update or similar.

Setting queues to be mirrored results in the action being persisted as a status field:

```
$ oc get rabbitmq.rabbitmq.openstack.org/rabbitmq -o json |jq .status.queueType
Mirrored
```

Jira: https://issues.redhat.com/browse/OSPRH-10790


controllers/rabbitmq/rabbitmq_controller.go
```diff
@@ -356,6 +363,31 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ct
 
 	if clusterReady {
 		instance.Status.Conditions.MarkTrue(condition.DeploymentReadyCondition, condition.DeploymentReadyMessage)
+		// Let's wait DeploymentReadyCondition=True to apply the policy
+		queue := instance.Spec.QueueType
+		ismirrored := isMirroredPolicyApplied(ctx, helper, instance, r.config)
```
dciabrin, line 368: Do I get that right that with the default ti Mirrored (what we want), every reconciliation will log a new line with the ha-policy of the cluster? If so, does that mean if will also perform one call into the container for each reconciliation?

lmiccini:
recheck

controllers/rabbitmq/rabbitmq_controller.go
```diff
@@ -356,6 +362,31 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ct
...
+		queue := instance.Spec.QueueType
+		ismirrored := instance.Status.QueueType
+
+		if queue == "Mirrored" && *instance.Spec.Replicas > 1 {
+			if ismirrored != "Mirrored" {
+				Log.Info("ha-all policy not present. Applying.")
+				err := updateMirroredPolicy(ctx, helper, instance, r.config, true)
+				if err != nil {
+					Log.Error(err, "Could not apply policy")
+				} else {
```
stuggi, line 375: the else is not required, or is it. could just set `instance.Status.QueueType = "Mirrored"` if no error occurred

controllers/rabbitmq/rabbitmq_controller.go
```diff
@@ -356,6 +362,31 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ct
...
+					instance.Status.QueueType = "Mirrored"
+				}
+			}
+		} else {
+			if ismirrored == "Mirrored" {
+				Log.Info("Removing ha-all policy")
+				err := updateMirroredPolicy(ctx, helper, instance, r.config, false)
+				if err != nil {
+					Log.Error(err, "Could not remove policy")
+				} else {
```
stuggi, line 385: same the else is not required, or is it. could just set instance.Status.QueueType = "" if no error occured

lmiccini:
recheck

controllers/rabbitmq/rabbitmq_controller.go
```diff
@@ -356,6 +362,31 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ct
 
 	if clusterReady {
 		instance.Status.Conditions.MarkTrue(condition.DeploymentReadyCondition, condition.DeploymentReadyMessage)
+		// Let's wait DeploymentReadyCondition=True to apply the policy
+		queue := instance.Spec.QueueType
+		ismirrored := instance.Status.QueueType
+
+		if queue == "Mirrored" && *instance.Spec.Replicas > 1 {
+			if ismirrored != "Mirrored" {
```
stuggi, line 370: probably no need for a nested if. could be part of the above.

lmiccini: ouch right

controllers/rabbitmq/rabbitmq_controller.go
lmiccini: we can probably return err and get rid of the else?

stuggi: right, missed to read that and expected we return on err :) yes I think we should return on err and set a condition for it

controllers/rabbitmq/rabbitmq_controller.go
```diff
@@ -356,6 +362,39 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ct
...
+				instance.Status.Conditions.Set(condition.FalseCondition(
+					condition.DeploymentReadyCondition,
+					condition.ErrorReason,
+					condition.SeverityWarning,
+					condition.DeploymentReadyErrorMessage, err.Error()))
+				return ctrl.Result{}, err
+			}
+			instance.Status.QueueType = "Mirrored"
+		} else {
+			if ismirrored == "Mirrored" {
```
stuggi, line 383: `else if queue != Mirrored && ismirrored == "Mirrored"` ?

lmiccini:
recheck

stuggi (approved):
/lgtm

PR #389, opened 2025-05-06
Leader election lease tunables

hjensas:
Expose the leader election tunables for lease duration, renew deadline and retry period via environment variables.

This is inspired by rabbitmq operator exposes the same.

Jira: [OSPRH-16335](https://issues.redhat.com//browse/OSPRH-16335)

Depends-On: openstack-k8s-operators/lib-common#627


hjensas:
/test functional

I am seeing the same errors locally, but only every 3-4 times I run the functional tests.

stuggi:
I have seen it too this week, but could not spot the root cause for it, yet. I plan to look into it later this week.

PR #393, opened 2025-05-09
[DNM] replace redis with valkey

raukadah:
Note: It is added for testing purpose


lmiccini:
we can close this since we merged a pr to support both redis and valkey.

PR #394, opened 2025-05-09
[bgp] fix Makefile test target + add envtest test case

stuggi:
Adds bgp pkg to run functional tests with `make test`. Also adds an envtest to validate when BGPConfiguration CR is deleted, finalizer logic removes owned FRRConfiguration objects.


abays (approved):
/lgtm

lmiccini (approved):

PR #398, opened 2025-05-13
Append to AdditionalConfig instead of overriding

lmiccini:
This commit changes the way we enforce rabbitmq additionalConfig from immutable string with hard coded values to allow customers to append their own modifications.

Jira: https://issues.redhat.com/browse/OSPRH-16586


olliewalsh:
looks good to me, but lets wait for Damien to review

PR #403, opened 2025-06-16
Optimize pkg/ipam

dprince:
Significantly reduce net/netip.Addr.string4 allocations by

 * caching exclude and reservation IPs
 * reduce inner loop conversion to strings
 * reduce inner loop parsing of strings and comparison

Jira: [OSPRH-17235](https://issues.redhat.com//browse/OSPRH-17235)

### DIFF
--- a/controllers/network/ipset_controller.go
+++ b/controllers/network/ipset_controller.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"net/netip"
 	"sort"
 	"strings"
 
@@ -432,8 +433,11 @@ func (r *IPSetReconciler) ensureReservation(
 		}
 
 		if ipsetNet.FixedIP != nil {
-			ipDetails.FixedIP = net.ParseIP(string(*ipsetNet.FixedIP))
-			if ipDetails.FixedIP == nil {
+			ipDetails.FixedIP, err = netip.ParseAddr(string(*ipsetNet.FixedIP))
+			if err != nil {
+				return nil, fmt.Errorf("failed parse FixedIP %s", string(*ipsetNet.FixedIP))
+			}
+			if !ipDetails.FixedIP.IsValid() {
 				return nil, fmt.Errorf("failed parse FixedIP %s", string(*ipsetNet.FixedIP))
 			}
 		}
--- a/pkg/ipam/funcs.go
+++ b/pkg/ipam/funcs.go
@@ -2,11 +2,9 @@ package ipam
 
 import (
 	"fmt"
-	"net"
 	"net/netip"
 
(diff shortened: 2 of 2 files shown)


dprince:
recheck

controllers/network/ipset_controller.go
```diff
@@ -432,8 +433,8 @@ func (r *IPSetReconciler) ensureReservation(
 		}
 
 		if ipsetNet.FixedIP != nil {
-			ipDetails.FixedIP = net.ParseIP(string(*ipsetNet.FixedIP))
-			if ipDetails.FixedIP == nil {
+			ipDetails.FixedIP = netip.MustParseAddr(string(*ipsetNet.FixedIP))
```
stuggi, line 436: `MustParseAddr` panics on error, is this intended to be used? thats why we used the ParseIP version which does not panic

pkg/ipam/funcs.go
```diff
@@ -15,12 +13,45 @@ type AssignIPDetails struct {
...
+	reserved map[netip.Addr]string
+}
+
+func (a *AssignIPDetails) buildExcluded() {
+	if a.excluded != nil {
+		return
+	}
+	a.excluded = make(map[netip.Addr]bool, len(a.SubNet.ExcludeAddresses))
+	for _, ipStr := range a.SubNet.ExcludeAddresses {
+		if ip, err := netip.ParseAddr(ipStr); err == nil {
```
stuggi, line 28: We validate the ExcludeAddresses via the webhook, but not using netip. https://github.com/openstack-k8s-operators/infra-operator/blob/main/apis/network/v1beta1/netconfig_webhook.go#L345 
Should there be a validation here, too? or is it ok to just ignore a ParseAddr error?

pkg/ipam/funcs.go
```diff
@@ -15,12 +13,45 @@ type AssignIPDetails struct {
...
+}
+
+func (a *AssignIPDetails) buildReserved() {
+	if a.reserved != nil {
+		return
+	}
+	a.reserved = make(map[netip.Addr]string, len(a.Reservelist.Items))
+	for _, r := range a.Reservelist.Items {
+		if res, ok := r.Spec.Reservation[a.NetName]; ok {
+			if ip, err := netip.ParseAddr(res.Address); err == nil {
```
stuggi, line 41: same as above, ok to ignore err?

controllers/network/ipset_controller.go
dprince: Will switch to ParseAddr... MustParseAddr should only be used for tests :)

pkg/ipam/funcs.go
dprince: good call. fixed

stuggi (approved):
/lgtm

PR #405, opened 2025-06-19
Update to rabbitmq cluster-operator v2.11.0

steveb:
This is the version which other operators have when this is an indirect dependency [1].

[1] https://github.com/openstack-k8s-operators/openstack-operator/blob/main/apis/go.mod#L86

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,7 +13,7 @@ require (
 	github.com/openstack-k8s-operators/infra-operator/apis v0.6.1-0.20250319162810-463dd75a4cc4
 	github.com/openstack-k8s-operators/lib-common/modules/common v0.6.1-0.20250508141203-be026d3164f7
 	github.com/openstack-k8s-operators/lib-common/modules/test v0.6.1-0.20250508141203-be026d3164f7
-	github.com/rabbitmq/cluster-operator/v2 v2.9.0
+	github.com/rabbitmq/cluster-operator/v2 v2.11.0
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -13,7 +13,7 @@ require (
 	github.com/openstack-k8s-operators/infra-operator/apis v0.6.1-0.20250319162810-463dd75a4cc4
 	github.com/openstack-k8s-operators/lib-common/modules/common v0.6.1-0.20250508141203-be026d3164f7
 	github.com/openstack-k8s-operators/lib-common/modules/test v0.6.1-0.20250508141203-be026d3164f7
-	github.com/rabbitmq/cluster-operator/v2 v2.9.0
+	github.com/rabbitmq/cluster-operator/v2 v2.11.0
```
olliewalsh, line 16: This should be v2.9.0, we use a fork of the upstream operator. Should be fixed in whichever operator depends on the wrong version

olliewalsh: https://github.com/openstack-k8s-operators/openstack-operator/pull/1518

stuggi: we also have https://github.com/openstack-k8s-operators/renovate-config/blob/main/default.json5#L155 to keep it bellow 2.10. a1dc0fb16970a819eb1e7c818754c4c280b0a5d5 in the openstack-operator which bumped it to 2.11 was landed before that.

stuggi:
closing this PR, we recently bumped to v2.16.0

PR #424, opened 2025-08-11
Make fencing details fetching agnostic to short vs fqdn hostname

lmiccini:
By switching the comparison we make sure we match vs both short hostnames and fqdns when fetching fencing details from the secret.

Jira: https://issues.redhat.com/browse/OSPRH-18993


lmiccini:
recheck

PR #429, opened 2025-08-18
Extend queueType to Quorum

lmiccini:
This commit adds Quorum as a valid option for QueueType.
We will switch from the previous default (Mirrored) over to this new value when all the operators will be ready.

The implementation works by adding a new key into the transportURL "quorumqueues"="true" if the rabbitmq instance has been configured for Quorum Queues, so that every operator could read this and configure the respective services to use quorum queues.
It also adds a queuetype status field to signal the secret change.

Jira: https://issues.redhat.com/browse/OSPRH-19160

### DIFF
--- a/apis/bases/rabbitmq.openstack.org_rabbitmqs.yaml
+++ b/apis/bases/rabbitmq.openstack.org_rabbitmqs.yaml
@@ -1361,11 +1361,12 @@ spec:
                 type: object
               queueType:
                 default: Mirrored
-                description: QueueType to eventually apply the ha-all policy to the
-                  cluster
+                description: QueueType to eventually apply the ha-all policy or configure
+                  default queue type for the cluster
                 enum:
                 - None
                 - Mirrored
+                - Quorum
                 type: string
               rabbitmq:
                 description: Configuration options for RabbitMQ Pods created in the
--- a/apis/bases/rabbitmq.openstack.org_transporturls.yaml
+++ b/apis/bases/rabbitmq.openstack.org_transporturls.yaml
@@ -108,6 +108,10 @@ spec:
                   the opentack-operator in the top-level CR (e.g. the ContainerImage)
                 format: int64
                 type: integer
+              queueType:
+                description: QueueType - the queue type from the associated RabbitMq
+                  instance
+                type: string
               secretName:
                 description: SecretName - name of the secret containing the rabbitmq
                   transport URL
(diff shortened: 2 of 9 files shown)


lmiccini:
recheck

apis/bases/rabbitmq.openstack.org_rabbitmqs.yaml
```diff
@@ -1360,12 +1360,13 @@ spec:
                     type: string
                 type: object
               queueType:
-                default: Mirrored
```
paramite, line 1363: Do we really want to change the default value? What will happen with already deployed clusters. Is a migration from mirrored to quorum queues straight?

SeanMooney: it depend on how this default is applied

we should not in this change for exsiting CRs

you cant actully change this in geenral without also doing a deployment of the dataplane nodes so that nova-comptue adn the ovn metadata proxy? are also aligned.

on the nova side only the cell rabbits are exposed to the compute nodes
so you could upgrade each cell at a time but you still need to do a datapalne deployment for each node set in the cell to restore connectivity


fro greenfhiled its ok to change the defualt to Quorum queues.
but there will need to be a docuemnted procedure that is done as a post upgreade step in the saem or a seperate mataince window to migrate teh existing CRs.

it would be good to ensure there is testing of this upgrade senario.

SeanMooney: what im thinking is it would be good to have a kuttl test that deploy wiht mirrored then updates to Quorum but to test ti properly we would need a updates edpm/tempst test that deploy with mirrored and runs tempest smoke, then upgrades teh operators, modifes the contolplane cr and does a dataplane deployment before runitn tempest a second time to confirm it continues to work.

you can do some degree of testing with evntest and kuttle but really somehting like this need to betested somewhere end to end with tempest.

lmiccini: Changing the default here will not change existing crs.
Migrating from classic queues to quorum queues is a disruptive change that requires either an in place rabbitmq queues/exchanges/etc wipe and redeclare or some other orchestrated approach. It is out of scope for now. 
We will provide a documented migration plan if people want to jump on this ship before we have a shared (and agreed upon) automated migration path for all the operators.

PR #430, opened 2025-08-19
Advertise /128 when using ipv6

lmiccini:
This commit extends the GetFRRPodPrefixes function to detect if pods have ipv4 or ipv6 addresses and use the proper prefix.

Jira: https://issues.redhat.com/browse/OSPRH-19214


stuggi (approved):
/lgtm

PR #432, opened 2025-08-21
Remove required on inline RabbitmqClusterSpecCore

stuggi:
This is a pre-step to upgrade to controller-tools. When updating to controller-tools v0.18.0 the required tag results getting an empty string added to the required list:

```
                 type: object
             required:
+            - ""
             - containerImage
             type: object
           status:
```

Lets remove the tag as it should not be needed when inline the RabbitmqClusterSpecCore.

Related: [OSPRH-12935](https://issues.redhat.com//browse/OSPRH-12935)


abays (approved):
/lgtm

stuggi:
recheck

PR #438, opened 2025-09-13
Fix ip reassignment for ipset after scale-in

rabi:
After scale-in reserved ips are released back to the pool and if the ipset reconciles for some reason i.e update or infra-operator pod deleted, then the released ip would be assigned to the an exisitng ipset which already has a reservation.
This changes the logic to not try reserve for ipset/network combination which has a reserved ip.

jira: https://issues.redhat.com/browse/OSPRH-19871

### DIFF
--- a/apis/network/v1beta1/ipset_webhook.go
+++ b/apis/network/v1beta1/ipset_webhook.go
@@ -82,7 +82,7 @@ func (r *IPSet) ValidateCreate() (admission.Warnings, error) {
 	basePath := field.NewPath("spec")
 
 	// validate requested networks exist in netcfg
-	allErrs = append(allErrs, valiateIPSetNetwork(r.Spec.Networks, basePath, &netcfg.Spec)...)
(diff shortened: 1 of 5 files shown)


apis/network/v1beta1/ipset_webhook.go
```diff
@@ -160,7 +160,7 @@ func (r *IPSet) ValidateDelete() (admission.Warnings, error) {
 // - FixedIP has correct IP version of subnet
 // - FixedIP is in the subnet cidr
 // - Route is only specified on a single network per IPFamily
-func valiateIPSetNetwork(
+func validateIPSetNetwork(
```
karelyatin, line 163: can change name in description too.

apis/network/v1beta1/ipset_webhook.go
```diff
@@ -251,7 +251,7 @@ func valiateIPSetNetwork(
 // - if subnet changed within a network
 // - if fixedIP changed
 // - if defaultRoute changed
-func valiateIPSetChanged(
+func validateIPSetChanged(
```
karelyatin, line 254: same here above

PR #440, opened 2025-09-15
Fix ip reassignment for ipset after scale-in

rabi:
After scale-in reserved ips are released back to the pool and if the ipset reconciles for some reason i.e update or infra-operator pod deleted, then the released ip would be assigned to the an exisitng ipset which has already a reservation.

jira: https://issues.redhat.com/browse/OSPRH-19871


abays (approved):
/lgtm

karelyatin:
/override rdoproject.org/github-check
unrelated container-selinux issue

PR #442, opened 2025-09-15
Switch to Quorum Queues by default

lmiccini:
This should wait until all the other operators patches are merged + bumps.

Related: https://issues.redhat.com/browse/OSPRH-19160


lmiccini:
recheck

PR #451, opened 2025-09-24
Switch DNSMasq to daemon mode to prevent TCP connection blocking

stuggi:
Remove --keep-in-foreground and --no-daemon flags to allow DNSMasq to run in daemon mode. This resolves issues where long-lived idle TCP connections (e.g., from nc or monitoring tools) block DNSMasq's single-threaded event loop, causing liveness probe timeouts and unnecessary pod restarts.

In daemon mode, DNSMasq forks a dedicated process for DNS serving with improved connection handling, while a monitoring script keeps the container alive by checking for the daemon process. Each pod could then handle up to 20 TCP connections, which is the hard limit of dnsmasq 2.85.

Changes:
- Remove --keep-in-foreground and --no-daemon flags
- Add daemon monitoring with pgrep check and sleep loop
- Maintain all existing DNSMasq configuration options

Fixes: Pod restarts when using long-lived TCP connections like 'nc -t'

Jira: OSPRH-20039

Co-authored-by: Claude (Anthropic) claude@anthropic.com


abays (approved):
/lgtm

stuggi:
recheck

PR #455, opened 2025-09-26
Force RabbitMQ to stick to TLSv1.2

dciabrin:
The combination of RabbitMQ/Erlang currently in use has issues with TLS1.3, which can cause disconnections/partitions in the rabbitmq cluster during TLS key update.

In order to mitigate this issue until the underlying issue is addressed, force RabbitMQ to use TLS1.2 for connections, in the same way as we did for OSP17 (see [OSPRH-15346](https://issues.redhat.com//browse/OSPRH-15346)).

Jira: https://issues.redhat.com/browse/OSPRH-20331


stuggi (approved):
/lgtm

PR #465, opened 2025-09-29
Drop to rabbit 2.16.0 api

dprince:
This drops to rabbit 2.16.0 to avoid the golang 1.24.6 dependency which we don't yet support downstream.

Jira: OSPRH-20205


apis/go.mod
```diff
@@ -70,18 +70,37 @@ require (
...
+replace k8s.io/apiextensions-apiserver => k8s.io/apiextensions-apiserver v0.31.13 //allow-merging
+
+replace k8s.io/cli-runtime => k8s.io/cli-runtime v0.31.13 //allow-merging
+
+replace k8s.io/code-generator => k8s.io/code-generator v0.31.13 //allow-merging
+
+replace k8s.io/component-base => k8s.io/component-base v0.31.13 //allow-merging
+
+// custom RabbitmqClusterSpecCore for OpenStackControlplane (v2.16.0_patches)
+replace github.com/rabbitmq/cluster-operator/v2 => github.com/openstack-k8s-operators/rabbitmq-cluster-operator/v2 v2.6.1-0.20250929174222-a0d328fa4dec //allow-merging
```
stuggi, line 102: just wondering if we can/should also bump the version of the user branch to 2.16.0, to make it more obvious that it is really that version and not 2.6.1

stuggi: we can follow up on this

PR #533, opened 2026-02-12
Add RabbitMQ federation support for multi-region deployments

xek:
## Summary
- Adds `RabbitMQFederation` CRD and controller to configure RabbitMQ exchange federation via the Management HTTP API
- Enables replicating messages (e.g. Keystone notifications) between RabbitMQ clusters in different regions, so services like `barbican-keystone-listener` can consume them from their local broker
- Extends the Management API client with federation upstream methods (`max-hops`, `prefetch-count`, `reconnect-delay` per the [RabbitMQ federation reference](https://www.rabbitmq.com/docs/federation-reference))

The upstream connection can be specified either by referencing a `RabbitmqCluster` CR in the same namespace (for testing) or via a Secret containing the AMQP URI (for cross-region deployments).

The controller:
- Creates a `federation-upstream` parameter on the local cluster
- Creates a policy matching the configured exchange pattern (`apply-to: exchanges`)
- Cleans up both on deletion

Closes: [OSPRH-25296](https://issues.redhat.com//browse/OSPRH-25296)
Assisted-by: Claude 4.6 Opus (Anthropic AI)

## Test plan
- [ ] Unit/functional tests for the new controller
- [ ] Deploy two RabbitMQ clusters, configure federation, verify exchange replication
- [ ] Test cross-region mode with `upstreamSecretRef`
- [ ] Verify cleanup on CR deletion


xek:
This approach is valid, but not considered at this point.

PR #538, opened 2026-02-26
Fix RabbitMQ webhook to preserve QueueType during operator upgrades

lmiccini:
Restore RabbitmqCluster existence check (removed in 0021016) to prevent defaulting QueueType on existing deployments during operator upgrades. This fixes PRECONDITION_FAILED errors when services redeclare queues with mismatched queue types.

The webhook now checks in order: user-specified value, existing CR spec, existing CR status, RabbitmqCluster existence, then defaults to Quorum only for truly new deployments.

Tested:
- update from fr1 -> preserved queueType=nil
- update from fr3 -> preserved queueType=Mirrored
- new deployment -> queueType: Quorum

### DIFF
--- a/apis/rabbitmq/v1beta1/rabbitmq_webhook.go
+++ b/apis/rabbitmq/v1beta1/rabbitmq_webhook.go
@@ -18,8 +18,10 @@ package v1beta1
 
 import (
 	"context"
+	"time"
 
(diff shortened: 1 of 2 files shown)


stuggi (approved):
/lgtm

PR #551, opened 2026-03-18
Drop rabbitmq-cluster-operator dependency and manage RabbitMQ directly

lmiccini:
Remove the dependency on the external rabbitmq-cluster-operator and have
the infra-operator manage RabbitMQ StatefulSets, Services, ConfigMaps,
and Secrets directly.

Core controller changes:
- Direct StatefulSet management with proper volume mounts, config
  generation, and TLS support (both client and inter-node)
- Service creation for client (AMQP/AMQPS) and headless node discovery
- ConfigMap generation for server config, plugins, and config-data
- Secret management for default-user credentials and Erlang cookie
- PodDisruptionBudget for multi-replica deployments
- Fix stale ownerReferences in volumeClaimTemplates from adopted
  StatefulSets (orphan-delete + recreate with annotation-based storage
  class preservation)
- Label pods with skipPreStopChecks before StatefulSet deletion so the 
  Downward API volume is populated when cascade deletion triggers the 
  PreStop hook, preventing 7-day termination hangs

Version upgrade workflow (3.x to 4.x):
- State machine with phases: None -> DeletingResources -> WaitingForCluster -> None
- Detects targetVersion changes and triggers storage wipe when crossing
  major versions (required by RabbitMQ for 3.x -> 4.x upgrades)
- Sets wipeReason=VersionUpgrade in status to track upgrade progress
- Deletes StatefulSet to stop all pods atomically, then recreates it
  with a wipe-data init container that clears /var/lib/rabbitmq on the 
  existing PVs (marker files prevent re-wipes across pod restarts)
- Tracks currentVersion in status after successful upgrade

Queue type migration (Mirrored to Quorum):
- Supports migrating from classic mirrored (ha-all policy) queues to
  quorum queues via spec.queueType change
- Triggers storage wipe with wipeReason=QueueTypeMigration
- Manages ha-all policy lifecycle: applies for Mirrored (replicas > 1), 
  removes when transitioning away from Mirrored
- Defaulting webhook forces queueType from Mirrored to Quorum when
  targetVersion is 4.x+, since mirrored queues are not supported in
  RabbitMQ 4.x. This enables the openstack-operator to upgrade from
  3.x (Mirrored) to 4.x and have the migration handled automatically
- Validation webhook rejects Mirrored+4.x as a safety net after
  defaulting

AMQP proxy sidecar:
- Python-based TCP proxy injected as a sidecar container when
  status.proxyRequired is true (after version upgrade or queue migration)
- Rewrites AMQP queue.declare frames to force durable=True and 
  x-queue-type=quorum, and exchange.declare frames to force durable=True
- Listens on port 5672 (plain) or 5671 (TLS) depending on TLS config
- Forwards connections to RabbitMQ backend on port 5673
- Removed via clients-reconfigured annotation once consumers reconnect
- Includes liveness/readiness probes and TLS certificate mounting

Migration from rabbitmq-cluster-operator:
- Detects migration by checking for an existing RabbitmqCluster CR with
  the same name; if none is found (or the CRD is not installed), the 
  controller skips all migration logic and sets OldCRCleaned=True,
  allowing both operators to run side-by-side managing their own 
  independent resources without conflict
- Adoption logic reparents existing StatefulSets, Services, and Secrets
  from old RabbitmqCluster owner to new RabbitMq CR
- Strips old ownerReferences from PVCs before deleting old CR to prevent
  cascade garbage collection
- Cleans up old RabbitmqCluster CR after successful adoption
- Fixes stale volumeClaimTemplate ownerReferences that cause new PVCs
  to be garbage-collected when scaling up adopted StatefulSets

Testing:
- Functional tests (envtest): RabbitMQ controller reconciliation,
  RabbitMQPolicy lifecycle, TransportURL (plain, TLS, custom user/vhost,
  credential rotation), VCT fix, combined version+queue migration,
  proxy sidecar, operator coexistence (OldCRCleaned set when no old CR
  exists, unrelated RabbitmqCluster CRs not touched)
- Kuttl integration tests: basic cluster deployment, cluster resource
  ownership, credential rotation with cleanup-blocked finalizer,
  deletion with dependent resources, plugin enable/disable, policy
  enforcement via rabbitmqctl, queue migration (Mirrored to Quorum)
  with AMQP proxy rewrite verification (classic non-durable -> quorum
  durable), resource management (vhost/user/policy), scale-up with PDB,
  TLS configuration, TLS TransportURL, custom TransportURL, migration
  from old operator, version upgrades (3.9->4.2 with/without TLS,
  Mirrored upgrade), operator coexistence (both operators managing
  independent clusters without interference)

Depends-on: https://github.com/openstack-k8s-operators/openstack-operator/pull/1857


lmiccini:
retest

apis/rabbitmq/v1beta1/rabbitmq_types.go
```diff
@@ -69,47 +101,242 @@ type RabbitMqSpec struct {
...
+	// Certificate Authority's public certificate for TLS. The Secret must store this as ca.crt.
+	// This Secret can be created by running
+	// `kubectl create secret generic ca-secret --from-file=ca.crt=path/to/ca.crt`
+	// Used for mTLS, and TLS for rabbitmq_web_stomp and rabbitmq_web_mqtt.
+	CaSecretName string `json:"caSecretName,omitempty"`
+	// +kubebuilder:validation:Optional
+	// DisableNonTLSListeners - when set to true, disables non-TLS listeners for RabbitMQ,
+	// management plugin and for any enabled plugins in the following list: stomp, mqtt, web_stomp, web_mqtt.
+	// Only TLS-enabled clients will be able to connect.
+	DisableNonTLSListeners bool `json:"disableNonTLSListeners,omitempty"`
```
stuggi: I think we should not have omitempty the bool params. If a user sets it to false, omitempty strips it, and it might stay true.

lmiccini:
recheck

stuggi (approved):

amoralej:
FTR, this seems to have broken the telemetry-operator which uses the rabbitmqv1.RabbitmqCluster CRs created under the hood https://github.com/openstack-k8s-operators/telemetry-operator/blob/2e5cb810e29db97a9036d6710bf8e0cc39f983e5/internal/controller/metricstorage_controller.go#L769C17-L769C43

The deployment ci job is not enabling MetricStorage in telemetry https://logserver.rdoproject.org/859/rdoproject.org/859776a774cb412a9db7c3ff47a24aaf/controller/ci-framework-data/logs/openstack-must-gather/quay-io-openstack-k8s-operators-openstack-must-gather-sha256-58cb6caac54870493f224cc37a6252a09ea9ccbbda2851efaaba72cc8a6f1339/namespaces/openstack/crs/

It may be good add a job deploying additional services in infra-operator?

PR #560, opened 2026-04-08
Add Node RBAC permissions for MicroShift IPv6 detection

hjensas:
The RabbitMQ controller uses ocp.FirstClusterNetworkIsIPv6() from lib-common which falls back to checking Node PodCIDRs on MicroShift when the OpenShift Network CRD is not available. Add the necessary RBAC permissions to list Nodes at cluster scope.

Assisted-By: Claude (claude-sonnet-4.5)

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -38,6 +38,13 @@ rules:
 - apiGroups:
   - ""
   resources:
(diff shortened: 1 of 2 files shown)


stuggi (approved):
/lgtm

PR #563, opened 2026-04-13
Fix redis split-brain after pod-0 restart during failover

lmiccini:
When redis-redis-0 (the bootstrap pod) is deleted during a failover,
it restarts and tries to contact sentinel to find the current master.
Three problems caused it to fall through to the bootstrap path and
start a new independent master, creating a split-brain:

1. Single-try timeout: if sentinel was momentarily unreachable (e.g.
   the sentinel container on pod-0 itself was still starting), the
   3-second timeout expired and pod-0 immediately bootstrapped.

2. Headless service DNS: with PublishNotReadyAddresses: true, the
   headless service DNS can resolve to pod-0's own IP, so redis-cli
   connects to its own uninitialized sentinel instead of a peer.

3. Stale master identity: even when contacting a peer sentinel, it
   may still report the restarting pod as master (within the
   down-after-milliseconds window before failover completes).

Fix by adding a wait_for_master() function in common.sh that:
- Contacts each peer pod individually by FQDN (skipping self)
- Retries up to 10 times (30s total) before allowing bootstrap
- Rejects answers where the peer still thinks we are master

Also increase InitialDelaySeconds to 40s on all redis and sentinel
probes so Kubernetes doesn't kill the pod before the retry loop
completes, and remove unused TCP probe variables that were never
referenced by the redis container.

### DIFF
--- a/internal/redis/statefulset.go
+++ b/internal/redis/statefulset.go
@@ -27,36 +27,15 @@ func StatefulSet(
 	}
 	ls := labels.GetLabels(r, "redis", matchls)
 
-	livenessProbe := &corev1.Probe{
-		// TODO might need tuning
-		TimeoutSeconds:      5,
-		PeriodSeconds:       3,
-		InitialDelaySeconds: 3,
-	}
-	readinessProbe := &corev1.Probe{
-		// TODO might need tuning
-		TimeoutSeconds:      5,
-		PeriodSeconds:       5,
-		InitialDelaySeconds: 5,
-	}
 	sentinelLivenessProbe := &corev1.Probe{
-		// TODO might need tuning
 		TimeoutSeconds:      5,
 		PeriodSeconds:       3,
-		InitialDelaySeconds: 3,
+		InitialDelaySeconds: 40,
 	}
 	sentinelReadinessProbe := &corev1.Probe{
-		// TODO might need tuning
 		TimeoutSeconds:      5,
 		PeriodSeconds:       5,
-		InitialDelaySeconds: 5,
-	}
-
-	livenessProbe.TCPSocket = &corev1.TCPSocketAction{
-		Port: intstr.IntOrString{Type: intstr.Int, IntVal: int32(6379)},
-	}
-	readinessProbe.TCPSocket = &corev1.TCPSocketAction{
-		Port: intstr.IntOrString{Type: intstr.Int, IntVal: int32(6379)},
+		InitialDelaySeconds: 40,
 	}
 	sentinelLivenessProbe.TCPSocket = &corev1.TCPSocketAction{
 		Port: intstr.IntOrString{Type: intstr.Int, IntVal: int32(26379)},
@@ -78,6 +57,9 @@ func StatefulSet(
 	}, {
 		Name:  "CONFIG_HASH",
 		Value: configHash,
+	}, {
+		Name:  "REPLICAS",
+		Value: strconv.Itoa(int(*r.Spec.Replicas)),
 	}}
 
 	sts := &appsv1.StatefulSet{
@@ -115,6 +97,7 @@ func StatefulSet(
 										Command: []string{"/var/lib/operator-scripts/redis_probe.sh", "liveness"},
 									},
 								},
+								InitialDelaySeconds: 40,
 							},
 							ReadinessProbe: &corev1.Probe{
 								ProbeHandler: corev1.ProbeHandler{
@@ -122,6 +105,7 @@ func StatefulSet(
 										Command: []string{"/var/lib/operator-scripts/redis_probe.sh", "readiness"},
 									},
 								},
+								InitialDelaySeconds: 40,
 							},
 						}, {
 							Image:   r.Spec.ContainerImage,
--- a/templates/redis/bin/common.sh
+++ b/templates/redis/bin/common.sh
@@ -80,8 +80,77 @@ function remove_pod_label() {
     local pod="$1"
     local label="$2"
     local patch="[{\"op\": \"remove\", \"path\": \"/metadata/labels/${label}\"}]"
-    # 200: OK, 422: not found
-    configure_pod_label $pod "$patch" "(200|422)"
+    # 200: OK, 404: pod not found, 422: label not found
+    configure_pod_label $pod "$patch" "(200|404|422)"
+}
+
+# Wait for a peer sentinel to report a valid master for the cluster.
+# Contacts each peer pod individually by FQDN (skipping self) to avoid
+# the headless service DNS resolving to our own uninitialized sentinel.
+# If a peer still reports US as master (stale info before
+# down-after-milliseconds triggers failover), keeps retrying until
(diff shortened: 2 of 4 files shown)


lmiccini:
recheck

templates/redis/bin/common.sh
```diff
@@ -84,6 +84,44 @@ function remove_pod_label() {
...
+# Prints the master address on success (FQDN or IP).
+function wait_for_master() {
+    local retries=${SENTINEL_RETRIES:-10}
+    local delay=${SENTINEL_RETRY_DELAY:-3}
+    local pod_ordinal=${POD_NAME##*-}
+    local pod_base=${POD_NAME%-*}
+
+    for i in $(seq 1 $retries); do
+        local ordinal=0
+        while [ $ordinal -le 9 ]; do
```
dciabrin: This feels arbitrary and we are likely not supporting a 9-node redis deployment anyway.

templates/redis/bin/common.sh
```diff
@@ -84,6 +84,44 @@ function remove_pod_label() {
...
+            if [ "$ordinal" != "$pod_ordinal" ]; then
+                local peer="${pod_base}-${ordinal}.${SVC_FQDN}"
+                local output
+                output=$(timeout ${TIMEOUT} $REDIS_CLI_CMD -h ${peer} -p 26379 sentinel master redis 2>/dev/null) && {
```
dciabrin: This statement feels like a micro-optimization, and I believe this will break if we use set -e, as the logical test is performed outside of a if condition.

templates/redis/bin/common.sh
```diff
@@ -84,6 +84,44 @@ function remove_pod_label() {
...
+                    # pre-failover info — try remaining peers before waiting.
+                    if ! echo "$master" | grep -q "^${POD_NAME}\."; then
+                        echo "$master"
+                        return 0
+                    fi
+                }
+            fi
+            ordinal=$((ordinal + 1))
+        done
+        log "Attempt $i/$retries: no valid master found, retrying in ${delay}s..."
```
dciabrin: I'm not sure what would happen at the very first start? It looks like all pods would retry for a long time because there's no master sentinel yet?

dciabrin (changes requested):

templates/redis/bin/common.sh
lmiccini: +1, changed to return if no peers are reachable.

templates/redis/bin/common.sh
```diff
@@ -84,6 +84,52 @@ function remove_pod_label() {
     configure_pod_label $pod "$patch" "(200|422)"
 }
 
+# Wait for a peer sentinel to report a valid master for the cluster.
+# Contacts each peer pod individually by FQDN (skipping self) to avoid
+# the headless service DNS resolving to our own uninitialized sentinel.
+# If a peer still reports US as master (stale info before
+# down-after-milliseconds triggers failover), keeps retrying until
+# failover completes and a different master is elected.
+# If no peers are reachable at all (first deployment), returns
```
dciabrin: I get the improvement to not wait, but I have the feeling this suffers from the same point as the previous revision.
On the initial deployement, as long as only one pod is started, this will be the master. However if for a weird reason, all pod restart after some run (e.g. you forcibly stop redis on all pod at once), 3 pods will restart concurrently, so all pod will determine that no sentinel is running, and they will start in master, leaving a split brain.
Did I get that right, or am I missing the point entirely?

lmiccini: as far as I understood it it should work like:
  1. All 3 pods restart, all call wait_for_master
  2. No peers are reachable → all return failure immediately                                   
  3. Pod-0: is_bootstrap_pod → true → bootstraps as master  
  4. Pod-1, Pod-2: is_bootstrap_pod → false → exit 1, restart                                  
  5. On their next restart, pod-0's sentinel is up → they join as replicas

dciabrin: oh I missed the is_bootstrap_pod. OK this makes sense to me now, thanks.

internal/redis/statefulset.go
```diff
@@ -27,36 +27,15 @@ func StatefulSet(
...
 		TimeoutSeconds:      5,
 		PeriodSeconds:       3,
-		InitialDelaySeconds: 3,
+		InitialDelaySeconds: 40,
 	}
 	sentinelReadinessProbe := &corev1.Probe{
-		// TODO might need tuning
 		TimeoutSeconds:      5,
 		PeriodSeconds:       5,
-		InitialDelaySeconds: 5,
```
dciabrin, line 52: I think we want to keep the probes for the main container, if only for the specific TCP ports 6379?

lmiccini: I am just removing these unused variables, the actual probe is left intact (it is using the inline exec-based probes calling redis_probe.sh)

lmiccini:
recheck

lmiccini:
@coderabbitai full review

lmiccini:
@coderabbitai full review

lmiccini:
@coderabbitai full review

internal/controller/redis/redis_controller.go
```diff
@@ -510,6 +523,38 @@ func (r *Reconciler) generateConfigMaps(
...
+func (r *Reconciler) reconcileClusterState(
+	ctx context.Context,
+	instance *redisv1.Redis,
+	readyReplicas int32,
+) error {
+	// Authorize bootstrap only when no pod is Ready. This covers a fresh
+	// deployment and a full outage (all data is ephemeral, so a full outage is
+	// effectively a fresh start), while refusing bootstrap during failovers,
+	// rolling restarts, or network partitions where a master may still serve.
+	bootstrapAuthorized := strconv.FormatBool(readyReplicas == 0)
```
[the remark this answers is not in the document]
lmiccini: Thanks — this is a fair characterization and we're accepting the residual risk for this design. Rationale:

Redis here is backed by ephemeral `emptyDir` storage and used as a cache-style store. That makes three goals mutually exclusive: (a) ephemeral storage, (b) automatic recovery after a full outage, and (c) a hard, durable split-brain fence. A "set-true-exactly-once" fencing token (PVC or Lease) satisfies (c) but breaks (b), since a legitimate full-outage restart is indistinguishable from a fresh deploy without durable state.

Given that, `readyReplicas == 0` + operator-maintained ConfigMap is a deliberate soft fence, not a hard one. The remaining split-brain window requires the simultaneous coincidence of: a live-but-unreachable master (network partition), the operator observing 0 ready replicas, and pod-0 restarting inside the kubelet's ConfigMap-refresh lag. That is narrow, and bootstrap still requires `has_alive_peers` to return false, which catches the common running-but-unready peer case whenever pod-0 can reach it.

If we later move Redis to durable (PVC) storage, we'll replace this with a proper fencing token that is authoritative post-initialization. Tracking that as a follow-up rather than blocking this fix, which already removes the unconditional-bootstrap path that existed at the merge base.

PR #565, opened 2026-04-14
Reject RabbitMQUser when username matches cluster default user

lmiccini:
Add a validation webhook check that prevents creating a RabbitMQUser whose username conflicts with the default_user of the referenced RabbitMQ cluster. The check looks up the <cluster>-default-user secret and compares usernames, gracefully skipping if the cluster doesn't exist yet.


stuggi (approved):
/lgtm

PR #566, opened 2026-04-15
[b/r] Label RabbitMQ default-user secret for backup/restore

stuggi:
Add backup restore labels to the default-user secret so it is restored from backup. The RabbitMQ controller already reuses existing credentials (only generates on first creation), so restoring this secret preserves RabbitMQ passwords across backup/restore cycles. This eliminates the manual RabbitMQ credential restore step.


lmiccini (approved):

stuggi:
recheck

PR #567, opened 2026-04-15
Expose Redis TLS status, sentinel endpoints, and client URL helpers

lmiccini:
Add TLSSupport and SentinelHosts fields to RedisStatus, populated by
the controller from the TLS spec and statefulset pod DNS names. Expose
helper methods on the Redis API type for consumers:

- GetRedisTLSSupport / GetRedisClientURL: direct Redis connection URL 
- GetSentinelHosts / GetRedisSentinelURL: tooz-compatible sentinel URL 
  with sentinel_fallback, ssl, sentinel_ssl and ssl_ca_certs support
- GetRedisByName: fetch a Redis CR by name/namespace

TLSSupport uses a string enum (True/False/"") instead of bool to comply
with the NoBools CRD validation policy. SentinelHosts carries the 
x-kubernetes-list-type=atomic marker required by the CRD schema checker.

Includes functional tests covering TLS status, sentinel host population,
and client/sentinel URL generation, plus shared test helpers.

### DIFF
--- a/apis/bases/redis.openstack.org_redises.yaml
+++ b/apis/bases/redis.openstack.org_redises.yaml
@@ -285,6 +285,20 @@ spec:
                   the opentack-operator in the top-level CR (e.g. the ContainerImage)
                 format: int64
                 type: integer
+              sentinelHosts:
+                description: SentinelHosts - List of sentinel endpoints in host:port
+                  format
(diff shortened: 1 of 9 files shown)


stuggi (approved):
/lgtm

PR #568, opened 2026-04-16
Fix rabbitmqadmin breakage from default_user_tags in .rabbitmqadmin.conf

lmiccini:
Only include default_user_tags.administrator=true in default_user.conf for RabbitMQ 3.x. On 4.x the sed that builds .rabbitmqadmin.conf transformed it into the unrecognised key username_tags.administrator, causing rabbitmqadmin to crash with an AttributeError.

Also fix the sed patterns to anchor on the full key assignment (s/^default_user = /username = /) so other keys sharing the default_user prefix are never corrupted.


stuggi (approved):
/lgtm

PR #571, opened 2026-04-29
Add quorum-based ClusterAvailableCondition for RabbitMQ availability

lmiccini:
Replace the raw ReadyCount > 0 check (checkClusterReadiness) with a quorum-aware ClusterAvailableCondition. The new condition is True when ReadyCount >= ceil(Replicas/2), ensuring the cluster has quorum before allowing TransportURL, Policy, Vhost, and User operations.
    
ClusterAvailableCondition is a sub-condition of ReadyCondition but never blocks it: quorum is always satisfied before all replicas are ready.
IsReady() still means "fully reconciled" while IsAvailable() means "can serve traffic".


TristanCacqueray:
recheck

lmiccini:
recheck

PR #577, opened 2026-05-05
Use shared singleton vhost/user CRs with per-consumer finalizers

lmiccini:
Multiple TransportURL CRs pointing to the same RabbitMQ vhost/user
(e.g., cinder and heat both using vhost "openstack") would each
create separate RabbitMQVhost/RabbitMQUser CRs. Deleting any one 
of them destroyed the RabbitMQ resource for all other consumers.

This replaces per-TransportURL CRs with shared singletons named
deterministically via CanonicalVhostName/CanonicalUserName. Each
consumer adds a per-TransportURL finalizer (turl.openstack.org/t-<name>)
so the RabbitMQ resource is only deleted when all consumers are gone.
The RabbitMQ cluster CR is set as owner for automatic GC on cluster
deletion.

Shared CR lifecycle:
- Orphaned user CRs are labeled (rabbitmq.openstack.org/orphaned)
  instead of deleted, keeping them reclaimable by new consumers.
  Auto-deletion only proceeds after admin removes the 
  cleanup-blocked finalizer.
- Vhost CRs show "Waiting for external finalizers" status when
  deletion is blocked by user finalizer cleanup.
- User controller cascades vhost CR deletion after its own cleanup.

Migration from legacy per-TransportURL CRs:
- Legacy CRs (with TransportURL owner refs) are cleaned up after
  canonical CRs are created.
- Credentials are pre-copied to the canonical secret to avoid
  password regeneration that would break existing connections.
- Webhook skips legacy CRs and deleting CRs in uniqueness checks.

Webhook hardening:
- Reject permission changes on shared users with active consumers.
- Enforce RabbitmqClusterName immutability on user, vhost, and policy.
- Validate RabbitMQPolicy Pattern as valid regex at admission.
- Remove duplicate +kubebuilder:webhook markers from apis/ package.
- Propagate admission context and add timeouts for k8s API calls.

API client hardening:
- Add context.Context to all RabbitMQ API client methods.
- Bound error response reads to 1MB and drain bodies on success.
- Track actual vhost/policy name in RabbitMQPolicyStatus so
  reconcileDelete targets the correct resource.

Controller fixes:
- Move ObservedGeneration to end of reconcileNormal (user, vhost,
  policy) so it is only set after all work succeeds.
- Propagate PatchInstance errors via named return instead of
  swallowing them.
- Guard per-consumer finalizer names against the Kubernetes
  63-char limit with SHA-256 hash truncation.
- Add periodic 5-minute requeue for long TransportURL names
  where watch-based reverse mapping fails.

Backup/restore:
- Add backup.openstack.org labels to auto-generated credential
  secrets so Velero includes them.
- Set restore-order to 10 so the secret is restored before the 
  RabbitMQUser CR (order 40).


lmiccini:
recheck

lmiccini:
retest

apis/rabbitmq/v1beta1/conditions.go
```diff
@@ -32,16 +35,52 @@ const (
...
-	// RabbitMQUserCleanupBlockedFinalizer - temporary finalizer to block automatic cleanup of RabbitMQUsers
-	// This finalizer prevents TransportURL from automatically deleting users during credential rotation.
-	// It must be manually removed by an operator/admin to allow cleanup to proceed.
-	// TODO: Replace with proper safe-to-delete logic, then remove this finalizer from existing users.
+	// TransportURLFinalizerPrefix - prefix for per-TransportURL finalizers on shared vhost/user CRs.
+	// Use TransportURLFinalizerFor() to build the full finalizer name safely.
+	TransportURLFinalizerPrefix = "turl.openstack.org/t-"
+
+	// maxFinalizerNameSegment is the Kubernetes limit for the name segment after "/"
+	maxFinalizerNameSegment = 63
```
stuggi, line 47: nit, could use validation.DNS1123LabelMaxLength like in https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/common/webhook/rfc.go#L38C16-L38C48

stuggi (approved):
/lgtm

PR #578, opened 2026-05-06
[dnsmasq] Default local=/<namespace>.svc/ to resolve queries locally

stuggi:
Set local=/<namespace>.svc/ by default in DNSMasq webhook so that queries for the openstack.svc domain are resolved from hosts files rather than forwarded to upstream DNS servers. This prevents delays and SERVFAIL errors when upstream DNS is slow at resolving AAAA records for openstack.svc.

The namespace is passed dynamically through the Default() chain so the domain is correct for any deployment namespace.

Users can override the default by specifying their own local option. Existing CRs get the default on upgrade via controller-triggered webhook (EnsureWebhookTrigger pattern from lib-common).

Closes: [OSPRH-29710](https://redhat.atlassian.net/browse/OSPRH-29710)

Depends-On: https://github.com/openstack-k8s-operators/openstack-operator/pull/1917


stuggi:
recheck

stuggi:
recheck

stuggi:
recheck

stuggi:
recheck

stuggi:
for some reason the dns CM is not part of the rdo job must-gather, but from the pod log we see the following message that confirms that the change was added:
`  dnsmasq[1]: using only locally-known addresses for domain openstack.svc   `

one note:
```
  DNSMasq now defaults `local=/<namespace>.svc/` so that queries for the
  OpenStack namespace domain (e.g. `openstack.svc`) are resolved locally
  from hosts files instead of being forwarded to upstream DNS servers.
  This fixes delays and SERVFAIL errors caused by slow upstream DNS
  resolution of AAAA records (OSPRH-29689).

  **Possible Impact**: Any service in the OpenStack namespace that EDPM nodes need
  to resolve must have a corresponding `DNSData` hosts entry. Queries for                                                                                                                    
  `*.<namespace>.svc` are no longer forwarded to upstream DNS. All
  OpenStack service endpoints are already registered via `DNSData` CRs by
  the openstack-operator, so standard deployments are unaffected.  Services
  matching `*.<namespace>.svc` that are currently resolved via upstream
  DNS servers will now need a local `DNSData` entry instead.
```

abays (approved):
/lgtm

PR #580, opened 2026-05-13
Automate orphaned user cleanup via NodeSet credential verification

lmiccini:
This PR implements:

1. Automate orphaned user cleanup via NodeSet credential verification

When a TransportURL's username rotates, the old RabbitMQUser CR retains
a per-consumer finalizer that blocks its deletion until an operator
manually removes it. 

This adds a two-phase release state machine that automatically removes
the finalizer once NodeSet deployments confirm the new credentials are 
fully rolled out:

Phase 1 - Wait for service operators (Nova, Neutron, etc.) to
regenerate their config secrets with the new TransportURL credentials.
Detected when AreSecretHashesInSync returns false.

Phase 2 - Wait for a NodeSet deployment to push the new configs to
compute nodes. Detected when AreSecretHashesInSync returns true again.

The NodeSet watch is conditional on CRD availability so the operator
starts cleanly on clusters without the dataplane CRD installed (e.g.
KUTTL test environments). When absent, pending releases fall back to
periodic polling via ConnectionCheckInterval.

2. Fix Service reconcile loop causing RabbitMQ pod restarts on operator restart

Replace full ServicePort slice assignment with name-based merging to
preserve server-defaulted fields (e.g. NodePort). The previous approach
caused controllerutil.CreateOrPatch to detect a diff on every reconcile,
triggering an infinite update loop on the Client Service that cascaded
to StatefulSet restarts via the Owns watch.


internal/controller/rabbitmq/rabbitmq_controller.go
```diff
@@ -1773,3 +1773,30 @@ func (r *Reconciler) findObjectsForSrc(ctx context.Context, src client.Object) [
 
 	return requests
 }
+
+// mergeServicePorts merges desired service ports into existing ports matched by
+// name, preserving server-defaulted fields (NodePort). When port counts differ
+// or a desired port name is not found, the existing slice is replaced entirely.
+func mergeServicePorts(existing *[]corev1.ServicePort, desired []corev1.ServicePort) {
```
stuggi, line 1780: looks to be not related to the commit, but not a major change, so good to have I think

internal/controller/rabbitmq/transporturl_controller.go
```diff
@@ -702,8 +781,17 @@ func (r *TransportURLReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 			builder.WithPredicates(predicate.ResourceVersionChangedPredicate{}),
-		).
-		Complete(r)
+		)
+
+	gk := schema.GroupKind{Group: edpm.NodeSetGVK.Group, Kind: edpm.NodeSetGVK.Kind}
+	if _, err := mgr.GetRESTMapper().RESTMapping(gk, edpm.NodeSetGVK.Version); err == nil {
+		b = b.Watches(
+			edpm.NewNodeSetObject(),
+			handler.EnqueueRequestsFromMapFunc(r.findTransportURLsWithPendingRelease),
```
stuggi, line 760: we probably should limit to `builder.WithPredicates(predicate.ResourceVersionChangedPredicate{}),` to limit reconciles on actual changes?

internal/controller/rabbitmq/transporturl_controller.go
```diff
@@ -606,6 +635,45 @@ func (r *TransportURLReconciler) reconcileDelete(ctx context.Context, instance *
...
+						labels = map[string]string{}
+					}
+					if labels[rabbitmqv1.RabbitMQUserOrphanedLabel] != "true" {
+						labels[rabbitmqv1.RabbitMQUserOrphanedLabel] = "true"
+						pendingUser.SetLabels(labels)
+						updated = true
+						Log.Info("Marking pending user as orphaned (no remaining consumers)", "name", pendingRef)
+					}
+				}
+			}
```
stuggi: we probably can extract this in a func as its the same as bellow and in releasePendingUser()?

apis/rabbitmq/v1beta1/transporturl_types.go
```diff
@@ -58,10 +58,25 @@ type TransportURLStatus struct {
...
+	// Stored in status so that spec changes can be compared against the previously reconciled value.
+	RabbitmqClusterName string `json:"rabbitmqClusterName,omitempty"`
+
 	// RabbitmqUserRef - the name of the RabbitMQUser CR that was created or referenced by this TransportURL.
 	// Empty if using default cluster admin credentials (no dedicated RabbitMQUser CR)
 	RabbitmqUserRef string `json:"rabbitmqUserRef,omitempty"`
 
+	// PendingUserRelease - the name of a previous RabbitMQUser CR whose finalizer
+	// removal was deferred during a credential rotation, pending deployment verification.
+	PendingUserRelease string `json:"pendingUserRelease,omitempty"`
```
stuggi: wondering if PendingUserRelease would be better called PreviousRabbitMQUserRef or something to align with current RabbitmqUserRef and easier to understand what it references?

apis/rabbitmq/v1beta1/transporturl_types.go
```diff
@@ -58,10 +58,25 @@ type TransportURLStatus struct {
...
+
+	// PendingUserDeploymentDetected - set to "true" when AreSecretHashesInSync
+	// returns false after a credential rotation, indicating that a deployment
+	// with new credentials has started and we are waiting for it to complete.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Enum="true";"false"
+	PendingUserDeploymentDetected string `json:"pendingUserDeploymentDetected,omitempty"`
```
stuggi: would it be easier to inverse it and call it `NodeSetSynced`, to reflect whether NodeSets have synced the new config?

like with the above it could look like:
```
  PreviousRabbitmqUserRef:  "rabbitmq-cluster-user-nova-old"
  NodeSetSynced:            "false"
```

apis/rabbitmq/v1beta1/transporturl_types.go
```diff
@@ -58,10 +58,25 @@ type TransportURLStatus struct {
...
 
+	// PendingUserRelease - the name of a previous RabbitMQUser CR whose finalizer
+	// removal was deferred during a credential rotation, pending deployment verification.
+	PendingUserRelease string `json:"pendingUserRelease,omitempty"`
+
+	// PendingUserDeploymentDetected - set to "true" when AreSecretHashesInSync
+	// returns false after a credential rotation, indicating that a deployment
+	// with new credentials has started and we are waiting for it to complete.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Enum="true";"false"
```
stuggi: in https://github.com/openstack-k8s-operators/infra-operator/pull/580/changes#diff-627c3e6f14aec9dd16405a764cf2e60199490e15a399bd3df182df26f00e999dR367 it gets set to "", so should this be added as allowed, or removed from status when "", or always set to either true or false?

test/functional/transporturl_controller_test.go
```diff
@@ -908,13 +908,11 @@ var _ = Describe("TransportURL controller", func() {
 			// Simulate user being ready
 			SimulateRabbitMQUserReady(oldUserCRName, "/")
 
-			// Verify old user exists and has both finalizers
+			// Verify old user exists and has per-consumer finalizer
```
stuggi, line 912: maybe we also want to add test for the transition case when there is a nodeset and rotation will run?
  - NodeSets exist with secretHashes in sync
  - Secret hashes go out of sync
  - Secret hashes come back in sync
  - double rotation probably nice to have

internal/controller/rabbitmq/transporturl_controller.go
```diff
@@ -792,43 +872,126 @@ func (r *TransportURLReconciler) deleteOrphanedCR(ctx context.Context, obj clien
...
-	if err := r.Get(ctx, types.NamespacedName{Name: name, Namespace: namespace}, user); err != nil {
-		if k8s_errors.IsNotFound(err) {
-			return nil
+	pendingRef := instance.Status.PreviousRabbitmqUserRef
+
+	// No NodeSets with secretHashes -> no computes affected, release immediately
+	haveNS, err := edpm.HaveNodeSets(ctx, r.Client, instance.Namespace)
+	if err != nil {
+		Log.Error(err, "Failed to check for nodesets")
+		return false, nil
```
stuggi, line 895: If HaveNodeSets return an error we reconcile in 5min, I guess thats fine, instead of returning an err and have an immediate reconcile triggered?

internal/controller/rabbitmq/transporturl_controller.go
```diff
@@ -792,43 +872,126 @@ func (r *TransportURLReconciler) deleteOrphanedCR(ctx context.Context, obj clien
...
+	}
+	if !haveNS {
+		return r.releasePendingUser(ctx, instance)
+	}
+
+	inSync, info, err := edpm.AreSecretHashesInSync(ctx, r.Client, instance.Namespace)
+	if err != nil {
+		Log.Error(err, "Failed to check nodeset sync status for pending release")
+		return false, nil
```
stuggi, line 904: same as for HaveNodeSets

stuggi (commented):
looks good to me, just wanted to double check on the behavior of this this two

internal/controller/rabbitmq/transporturl_controller.go
lmiccini: the idea was to "fail safe" and succeed eventually (while keeping the old credentials around) instead of retrying faster. if you prefer we can flip this

stuggi: I think for this use case it should be fine to have a delay if there is an error to fetch nodesets.

stuggi (approved):
/lgtm

PR #583, opened 2026-05-21
Use hash-based naming for canonical RabbitMQ user/vhost CRs

lmiccini:
The old naming convention (<cluster>-<vhost>-user-<username>) could exceed the 63-char Kubernetes label limit with long component names (e.g. rabbitmq-notifications + watcher-notifications = 71 chars).

Switch to <username>-<sha8(cluster+vhost)> for users and <vhostname>-<sha8(cluster)> for vhosts. The human-readable part comes first for kubectl readability; the SHA256 hash disambiguates the cluster/vhost combination. Full metadata remains in the CR spec/status.

### DIFF
--- a/apis/rabbitmq/v1beta1/canonical_names_test.go
+++ b/apis/rabbitmq/v1beta1/canonical_names_test.go
@@ -1,0 +1,158 @@
+/*
+Copyright 2026.
+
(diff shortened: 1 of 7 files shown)


stuggi (approved):
/lgtm

PR #584, opened 2026-05-21
InstanceHA refactoring and improvements

lmiccini:
Hardening and refactoring:
- Refactor evacuation workflow with orchestrated host-level processing, configurable retry count (EVACUATION_RETRIES), and standardized FAILED marker prefix
- Add per-aggregate failure threshold via instanceha:max_failures host aggregate metadata key
- Add host heartbeat mechanism for dual-channel failure detection
- Add Application Credential authentication support as an alternative to username/password
- Fix NoneType error when disabled_reason is None
- Remove duplicate log message on service initialization

Observability:
- Add Prometheus metrics exposition via /metrics endpoint using prometheus_client (counters, gauges, histograms)
- Add aggregate threshold metric and alert rule

API changes:
- Add InstanceHaHeartbeatPort field to InstanceHaSpec
- Add AuthSpec with ApplicationCredentialSecret field
- Add webhook validation for new fields

Testing:
- Add test suites for heartbeat detection, heartbeat scale, thread safety, orchestrated evacuation, aggregate threshold, Kubernetes events, and coverage gaps
- Add heartbeat benchmark tooling
- Refactor existing test fixtures into shared conftest

Documentation:
- Add operator guide, architecture deep-dive, heartbeat performance analysis, Prometheus monitoring guide, and application credentials guide

PR #586, opened 2026-05-26
Fix InstanceHA documentation inaccuracies and add MetalLB guide

lmiccini:
Architecture doc:
- THRESHOLD 0/100 behavior was backwards
- Deployment YAML rewritten to match actual code
- UDPSocketManager snippet updated to IPv6 dual-stack
- RBAC changed from ClusterRole to namespaced Role
- EVACUATION_RETRIES uses DEFAULT_EVACUATION_RETRIES constant

Guide doc:
- Clarified aggregate filtering is fail-closed
- Fixed FENCING_TIMEOUT range from 5-300 to 5-120
- Added stable IP for UDP listeners via MetalLB LoadBalancer Service

Other:
- Controller comment fix per PR #584 review (status not patched on panic)
- Heartbeat perf doc: recvmsg -> recvfrom
- CRD sample: added instanceHaHeartbeatPort comment

### DIFF
--- a/docs/instanceha_architecture.md
+++ b/docs/instanceha_architecture.md
@@ -5,8 +5,6 @@
 InstanceHA is a high-availability service for OpenStack that automatically detects and evacuates instances from failed compute nodes.
 
 **Version**: 2.5
-**Code Size**: 3,645 lines
-**Test Suite**: 656 tests across 17 test suites
 
 ## Table of Contents
 
@@ -134,7 +132,7 @@ _config_map: Dict[str, ConfigItem] = {
     'HASH_INTERVAL': ConfigItem('int', 60, 30, 300),
     'ORCHESTRATED_RESTART': ConfigItem('bool', False),
     'SKIP_SERVERS_WITH_NAME': ConfigItem('list', []),
-    'EVACUATION_RETRIES': ConfigItem('int', 5, 1, 20),
+    'EVACUATION_RETRIES': ConfigItem('int', DEFAULT_EVACUATION_RETRIES, 1, 20),  # DEFAULT_EVACUATION_RETRIES = 5
 }
 ```
 
@@ -297,16 +295,22 @@ def track_host_processing(service: 'InstanceHAService', hostname: str):
 **UDP Socket Management**:
 ```python
 class UDPSocketManager:
-    """Context manager for UDP socket with proper resource cleanup."""
+    """Context manager for UDP socket with proper resource cleanup.
+    Supports IPv4, IPv6, and dual-stack (::) binding."""
     def __init__(self, udp_ip, udp_port, label='UDP'):
         self.udp_ip = udp_ip
         self.udp_port = udp_port
         self.label = label
+        self.socket = None
 
     def __enter__(self):
-        self.socket = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+        info = socket.getaddrinfo(self.udp_ip or '::', self.udp_port,
+                                  type=socket.SOCK_DGRAM)[0]
+        self.socket = socket.socket(info[0], info[1])
+        if info[0] == socket.AF_INET6:
+            self.socket.setsockopt(socket.IPPROTO_IPV6, socket.IPV6_V6ONLY, 0)
         self.socket.settimeout(1.0)
-        self.socket.bind((self.udp_ip, self.udp_port))
+        self.socket.bind(info[4])
         return self.socket
 
     def __exit__(self, exc_type, exc_val, exc_tb):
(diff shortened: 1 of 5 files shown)


docs/instanceha_architecture.md
```diff
@@ -5,8 +5,8 @@
 InstanceHA is a high-availability service for OpenStack that automatically detects and evacuates instances from failed compute nodes.
 
 **Version**: 2.5
-**Code Size**: 3,645 lines
-**Test Suite**: 656 tests across 17 test suites
+**Code Size**: 3,647 lines
```
dciabrin: We should likely get rid of these useless metrics

docs/instanceha_architecture.md
```diff
@@ -1649,6 +1655,12 @@ This ensures orphaned hosts are not left powered off indefinitely after a pod cr
...
 
+### Known Limitations
+
+**Lock reconciliation scope**: The VM unlock sweep at startup only checks servers on hosts that are `forced_down AND state='down'`. This covers the realistic crash scenario (pod dies mid-evacuation while the host is still fenced). However, if the pod crashes after evacuation succeeds and post-recovery clears `forced_down`, but before the `finally` block unlocks the VMs, the locked VMs on their new (healthy) hosts would not be found by the reconciliation. This window is extremely narrow — the `finally` block in `_server_evacuate_future` runs per-server immediately after each evacuation completes, so the crash would need to occur between the evacuation completing and the `finally` executing. In practice, a SIGKILL (OOM, node crash) during concurrent evacuation would leave VMs on hosts that are still `forced_down`, which the current reconciliation handles correctly.
+
+A broader scan (checking all servers with `locked_reason == LOCK_REASON_EVACUATION` regardless of host state) would close this theoretical gap but requires listing all servers in the cloud at startup, which may be expensive on large deployments.
```
dciabrin, line 1660: That reads like a note to self rather than a doc point, no?

docs/instanceha_architecture.md
```diff
@@ -1828,7 +1840,7 @@ with concurrent.futures.ThreadPoolExecutor(max_workers=WORKERS) as executor:
 
 ### Test Statistics
 
-- **Total Tests**: 656 across 17 test suites
+- **Total Tests**: 677 across 19 test suites
```
dciabrin: Let's remove that?

docs/instanceha_architecture.md
```diff
@@ -2148,8 +2231,8 @@ config:
 ```
 
 **Notes**:
-- Value of `0` disables threshold checking
-- Value of `100` blocks all evacuations
+- Value of `0` blocks all evacuations (any failure percentage exceeds 0%)
+- Value of `100` disables threshold checking (nothing can exceed 100%)
```
dciabrin, line 2229: Just to double check, the semantics described in the doc got reversed, was the doc incorrect in the previous version and this one fixes it?

docs/instanceha_architecture.md
```diff
@@ -3004,8 +3087,8 @@ config:
 
 ## References
 
-- **Code**: `instanceha.py` (3,572 lines)
-- **Tests**: 656 tests across 17 test suites
+- **Code**: `instanceha.py` (3,647 lines)
+- **Tests**: 677 tests across 19 test suites
```
dciabrin: Useless/soon-to-be-out-of-date metrics

docs/instanceha_guide.md
```diff
@@ -525,7 +562,7 @@ config:
 | false | false | true | Only instances on hosts in tagged aggregates |
 | true | true | true | Tagged flavor OR tagged image, AND host must be in tagged aggregate |
 
-When all three filters are enabled but no resources carry the tag, all instances are evacuated (fail-open behavior ensures evacuation works out of the box before tagging is configured).
+When `TAGGED_FLAVORS` and/or `TAGGED_IMAGES` are enabled but no flavors or images carry the tag, all instances pass the flavor/image filter (fail-open behavior ensures evacuation works out of the box before tagging is configured). However, `TAGGED_AGGREGATES` is fail-closed: if no aggregates carry the evacuable tag, no hosts pass the aggregate filter and no evacuations occur.
```
dciabrin: nitpick: how about removing the parentheses? because the description of the fail-open behaviour seems as important as the fail-closed description later on.

docs/instanceha_guide.md
```diff
@@ -598,6 +635,53 @@ config:
...
+metadata:
+  name: openstack-edpm
+spec:
+  services:
+    - download-cache
+    - bootstrap
+    - configure-network
+    # ... existing services ...
+    - nova
+    - instanceha-monitoring    # <-- add after nova
```
dciabrin: Just curious, is the "after nova" just informative or is it mandatory?

docs/instanceha_architecture.md
lmiccini: correct, threshold=100 means 100% of the computes can fail.

docs/instanceha_architecture.md
lmiccini: This is important to highlight because users may see their vms locked and not understand why. In theory this is a niche corner case, in practice I think it is better to have it documented together with the reason why I decided not to implement that global scan.

docs/instanceha_guide.md
lmiccini: We could argue about this. Tagged_flavors and tagged_images are straight up 1:1 ports of the old instanceha implementation, so if you don't tag you can ignore them. 
Tagged_aggregates instead is the equivalent to the tripleo role-based assignment, where you actively added pacemakerremote to configure computes to be evacuable vs now that you just tag the aggregate, so imho this is more "intentional".

docs/instanceha_guide.md
lmiccini: order doesn't really matter, it was just to say "without nova the monitoring agent doesn't make much sense anyway". we can remove the comment if you prefer.

dciabrin (approved):

PR #587, opened 2026-05-27
OCP 4.20 frrconfiguration migration between namespaces

lmiccini:
OCP 4.20 changes the namespace for frrk8s from the metallb to a dedicated one.
This commit adds logic to migrate existing frrconfigurations to the new namespace.

We keep this simple:
  1. Every reconcile checks which namespace to use (GetFRRConfigurationNamespace) and creates/patches FRRConfigurations there via CreateOrPatch
  2. Every reconcile runs cleanupOldNamespaceFRRConfigurations when it detects the namespaces differ — this is a DeleteAllOf with label matching, which is a no-op if the old configs are already gone
  3. If the operator crashes mid-migration (e.g., after creating in the new namespace but before cleaning up the old), the next reconcile will simply retry the cleanup

### DIFF
--- a/internal/controller/network/bgpconfiguration_controller.go
+++ b/internal/controller/network/bgpconfiguration_controller.go
@@ -56,8 +56,9 @@ import (
 // BGPConfigurationReconciler reconciles a BGPConfiguration object
 type BGPConfigurationReconciler struct {
 	client.Client
-	Kclient kubernetes.Interface
-	Scheme  *runtime.Scheme
+	Kclient               kubernetes.Interface
+	Scheme                *runtime.Scheme
+	FRRMigrationNamespace string
 }
 
 // GetLogger returns a logger object with a prefix of "controller.name" and additional controller context fields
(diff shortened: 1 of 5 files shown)


eduolivares (approved):

internal/controller/network/bgpconfiguration_controller.go
```diff
@@ -368,6 +382,12 @@ func (r *BGPConfigurationReconciler) reconcileDelete(ctx context.Context, instan
 		return ctrl.Result{}, fmt.Errorf("error DeleteAllOf FRRConfiguration: %w", err)
 	}
 
+	if frrNamespace != instance.Spec.FRRConfigurationNamespace {
+		if err := r.cleanupOldNamespaceFRRConfigurations(ctx, instance, instance.Spec.FRRConfigurationNamespace); err != nil {
+			Log.Error(err, "Failed to cleanup FRRConfigurations from old namespace", "namespace", instance.Spec.FRRConfigurationNamespace)
```
abays, line 387: Maybe we should return the error here and kick out of the reconcile instead of logging and continuing.  L392 will remove the finalizer and then the CR is finally deleted, but if there was an error, there could be orphaned `FRRConfigurations`, perhaps?

PR #589, opened 2026-05-27
Fix FRR old namespace cleanup error handling in reconcileDelete

lmiccini:
- Return the error from `cleanupOldNamespaceFRRConfigurations` in `reconcileDelete` instead of just logging it
- Without this fix, if the cleanup fails, the finalizer is still removed and the CR is garbage collected — potentially leaving orphaned `FRRConfiguration` resources in the old namespace
- Addresses review feedback from #587 (https://github.com/openstack-k8s-operators/infra-operator/pull/587#discussion_r3311652379)


abays (approved):
/lgtm

PR #590, opened 2026-05-27
Add instanceha metrics service

lmiccini:
With this commit we add a Kubernetes Service exposing the InstanceHA metrics endpoint (port 8080/TCP) with labels "metrics: enabled" and "service: instanceha" for Prometheus scrapeconfig discovery by telemetry-operator.
We also add an optional MetricsTLS field (tls.SimpleService) to the InstanceHa API, allowing TLS certificate configuration for the metrics endpoint.
With https://github.com/openstack-k8s-operators/openstack-operator/pull/1932 we automatically create the certificate from openstack-operator when podlevel tls is enabled, so the instanceha controller checks if this exists (and the user did not override it via MetricsTLS) and automatically sets the proper field/adds the volume etc.


docs/instanceha_prometheus.md
```diff
@@ -17,6 +19,34 @@ The metrics are served by the `prometheus_client` Python library on the same HTT
 
 ---
 
+## TLS Configuration
+
+When `OpenStackControlPlane` has pod-level TLS enabled (`spec.tls.podLevel.enabled: true`), the openstack-operator automatically provisions a cert-manager Certificate for the InstanceHA metrics endpoint. This produces a Kubernetes TLS secret (`cert-instanceha-metrics`) containing `tls.crt`, `tls.key`, and `ca.crt`.
+
+The infra-operator InstanceHA controller **auto-detects** this secret: if the default secret `cert-instanceha-metrics` exists in the namespace, TLS is enabled automatically without any configuration on the InstanceHa CR. The controller:
```
dciabrin, line 26: I wonder if line 27-30 are just implementation details here. I would think it is sufficient to say "if TLS is enabled, metrics and probes use HTTPS".

docs/instanceha_prometheus.md
```diff
@@ -482,37 +518,23 @@ When the [telemetry-operator](https://github.com/openstack-k8s-operators/telemet
...
 | telemetry-operator (COO) | `prometheus-metric-storage` in `openstack` | `metric-storage-prometheus.openstack.svc:9090` |
 
-The PodMonitor approach described above places InstanceHA metrics in the OpenShift user workload Prometheus. If you want InstanceHA metrics alongside other OpenStack metrics (Ceilometer, RabbitMQ, node-exporter, OVN) in the COO Prometheus, create a `ScrapeConfig` CR instead.
+### Automatic Discovery (default)
 
-### Creating a ScrapeConfig for COO Prometheus
+The telemetry-operator **automatically discovers and scrapes InstanceHA metrics** — no manual configuration is required. The infra-operator creates a Kubernetes Service (`<instance-name>-metrics`) with the labels `metrics: enabled` and `service: instanceha`. The telemetry-operator's `MetricStorage` controller watches for Services with these labels and automatically generates a `ScrapeConfig` CR named `telemetry-instanceha` targeting port 8080.
 
-The COO Prometheus only picks up CRs with the label `service: metricStorage`. Create a `ScrapeConfig` targeting the InstanceHA pod:
+This works the same way as the OVN metrics integration. When a `MetricStorage` CR exists in the namespace:
```
dciabrin, line 525: I like this sentence, it brings this review into perspective.

templates/instanceha/bin/instanceha.py
```diff
@@ -1393,6 +1393,14 @@ def do_GET(self):
...
         try:
             server = HTTPServer(('', HEALTH_CHECK_PORT), HealthHandler)
+            tls_cert = os.getenv('METRICS_TLS_CERT')
+            tls_key = os.getenv('METRICS_TLS_KEY')
+            if tls_cert and tls_key:
+                import ssl
+                ssl_ctx = ssl.SSLContext(ssl.PROTOCOL_TLS_SERVER)
+                ssl_ctx.load_cert_chain(tls_cert, tls_key)
+                server.socket = ssl_ctx.wrap_socket(server.socket, server_side=True)
+                logging.info("Metrics endpoint serving over HTTPS on port %d", HEALTH_CHECK_PORT)
```
dciabrin, line 1403: We should add a log for the case we're serving HTTP.
It's not going to be the default, but I'm sure it will easy debugging.

docs/instanceha_prometheus.md
lmiccini: if you bring your own ca/crts and openstack-operator doesn't create the certificates for you then you would need to configure the metricsTLS field.

templates/instanceha/bin/instanceha.py
lmiccini: ack, will follow up in a subsequent review

dciabrin (approved):

PR #591, opened 2026-06-01
InstanceHA: security hardening, HMAC heartbeat authentication, and controller lifecycle fixes

lmiccini:
This PR hardens InstanceHA security, introduces authenticated heartbeat packets, and fixes controller lifecycle gaps.

  #### Security hardening & HMAC heartbeat (HBV2)

  - **HMAC-SHA256 heartbeat authentication**: Replace the unsigned HBV1 heartbeat
    protocol with HBV2, which adds per-packet HMAC-SHA256 signatures and timestamp
    freshness checks (120s replay window). Unsigned packets are rejected when HMAC
    keys are loaded. The HMAC key is auto-generated in an instance-scoped Secret
    with annotation-based rotation for zero-downtime key rollover.

  - **MetricsTLS hardening**: Add `minTLSVersion` (default 1.2) and `cipherSuites`
    controls to the InstanceHa API. 

  - **Credential protection**: Add JSON credential sanitization and minimize the
    IPMI subprocess environment to prevent credential leakage through logs or
    process listings.

  - **Instance scoping**: Scope the HMAC secret name and pod/service selector to
    the CR instance name, allowing multiple InstanceHa CRs to coexist in a
    namespace. Includes Deployment selector migration handling.

  - **Bug fixes**: Fix probe initialization semantics, thread
    shutdown_event through evacuation workers for clean SIGTERM handling, and add
    TLSInputReadyCondition to the initial condition list.

  #### Controller lifecycle

  - Add finalizer and `reconcileDelete` to the InstanceHA controller so that
    deleting an InstanceHa CR properly cleans up the Topology CR finalizer
    placed by `EnsureServiceTopology`.
    
  #### Minor

  - Normalize IPv4-mapped IPv6 addresses (`::ffff:` prefix) in UDP listener logs.


apis/instanceha/v1beta1/instanceha_types.go
```diff
@@ -46,6 +46,23 @@ type AuthSpec struct {
 	ApplicationCredentialSecret string `json:"applicationCredentialSecret,omitempty"`
 }
 
+// InstanceHaMetricsTLS extends tls.SimpleService with TLS hardening parameters.
+type InstanceHaMetricsTLS struct {
```
dciabrin, line 50: Will that require a dedicated doc or setup for FIPS?

apis/instanceha/v1beta1/instanceha_webhook.go
```diff
@@ -159,6 +167,33 @@ func (r *InstanceHa) validateUniqueOpenStackCloud(
 	return allErrs
 }
 
+// validateCipherSuites rejects cipher strings that would result in unencrypted connections.
```
dciabrin, line 170: Should we have that in lib-common instead? There may be a place that checks that for PQC or FIPS already?

lmiccini:
recheck

docs/instanceha_guide.md
```diff
@@ -633,24 +634,33 @@ config:
...
-kind: OpenStackDataPlaneService
-metadata:
-  name: instanceha-monitoring
-spec:
-  playbook: osp.edpm.instanceha_monitoring
+```bash
+kubectl annotate instanceha instanceha instanceha.openstack.org/rotate-hmac-key=true
+```
+
+This copies the current key to `hmac-key-previous`, generates a new key, and removes the annotation. The receiver accepts both keys during the rotation window. After redeploying compute nodes with the new key, the previous key can be cleared:
```
dciabrin, line 649: So the way we deal with current, previous secret in the mariadb operator is to have secret read-only, and a reference to the current secret in the status. if a new secret is to be created, the name of the secret is set in the spec file.

Could we do the same, in order to keep a consistenent way of managing secret rotation?

docs/instanceha_guide.md
```diff
@@ -633,24 +634,33 @@ config:
...
+
+```bash
+kubectl patch secret instanceha-heartbeat-hmac -p '{"data":{"hmac-key-previous":""}}'
```
dciabrin, line 652: We need a manual step to remove the old key from the secret once all the computes have been updated by edpm, correct?

internal/controller/instanceha/instanceha_controller.go
```diff
@@ -456,6 +472,104 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ct
...
+				return ctrl.Result{}, fmt.Errorf("failed to generate heartbeat HMAC key: %w", err)
+			}
+			hmacSecret = &corev1.Secret{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      heartbeatHMACSecretName,
+					Namespace: instance.Namespace,
+					Labels: labels.GetLabels(instance, labels.GetGroupLabel("instanceha"), map[string]string{
+						"backup.openstack.org/category":      "controlplane",
+						"backup.openstack.org/restore":       "true",
+						"backup.openstack.org/restore-order": "10",
```
dciabrin, line 492: Have we tested that this ordering works with the control plane b/r?

internal/instanceha/service.go
```diff
@@ -23,16 +23,16 @@ import (
 // MetricsService exposes the InstanceHA metrics endpoint for Prometheus scraping
 func MetricsService(instance *instancehav1.InstanceHa) *corev1.Service {
 	svcLabels := labels.GetLabels(instance, labels.GetGroupLabel("instanceha"), map[string]string{
-		common.AppSelector: "instanceha",
-		"metrics":          "enabled",
+		"service": "instanceha",
```
dciabrin, line 26: Why change this label? is it the new standard in lib-common?

apis/instanceha/v1beta1/instanceha_types.go
lmiccini: it is fips-compliant alraedy, takes the underlying ocp/container enforcement and restricts the available ciphers from the ones that you see here.

docs/instanceha_guide.md
lmiccini: This is different because we then have to glue things together and define a dataplaneservice referencing the secret name as datasource. If the secret changes it means that we would also have to change the service definition:
~~~
apiVersion: dataplane.openstack.org/v1beta1
kind: OpenStackDataPlaneService
metadata:
  name: instanceha-monitoring
spec:
  dataSources:
    - secretRef:
        name: instanceha-heartbeat-hmac
        optional: true
  playbook: osp.edpm.instanceha_monitoring
  edpmServiceType: instanceha-monitoring
~~~

docs/instanceha_guide.md
lmiccini: yes, I have patches to auto cleanup. Will send them after FR6.

internal/controller/instanceha/instanceha_controller.go
lmiccini: follows the same approach as rabbitmquser, where we restore the secret very early on so that users can be recreated using the same credentials.

internal/instanceha/service.go
lmiccini: This change is required because you can potentially have multiple instanceha copies per namespace, each  targeting a different cloud or region. With this change I leave the service: label required by telemetry-operator (so it generates the prometheus scrapeconfig to automatically pull the metrics) but lets us have a per-cr service endpoint, targeting the specific instanceha CR. This wouldn't be possible otherwise.

apis/instanceha/v1beta1/instanceha_webhook.go
lmiccini: AFAIK right now instanceha is the only service that uses this pattern (python tls endpoint), other services use httpd. I have in my todo list to move to lib-common if this pattern spreads, or we can switch to a lib-common helper if we add a check function.

lmiccini:
recheck

lmiccini:
recheck

internal/controller/instanceha/instanceha_controller.go
dciabrin: ack perfect

dciabrin (approved):

PR #594, opened 2026-06-05
Add network partition detection to prevent unsafe fencing

lmiccini:
Add two complementary mechanisms to prevent InstanceHA from fencing healthy hosts when the pod's worker node is network-isolated:

K8s API connectivity check:
- Background thread polls the K8s API every K8S_API_CHECK_INTERVAL seconds (default 15). After 3 consecutive failures, the agent marks itself not-ready and blocks all fencing until connectivity restores.
- Any non-5xx response is treated as reachable (covers 401/429/etc).
- Fencing gate runs before Nova API calls to avoid wasted requests on a partitioned network.
- Falls back to reading the ServiceAccount namespace file when POD_NAMESPACE env var is not set.

Heartbeat cliff detection:
- Tracks active heartbeat host count between poll cycles. If the count drops by more than HEARTBEAT_CLIFF_THRESHOLD% (default 50%) in a single cycle (minimum 3 hosts previously active), fencing is skipped.
- Baseline count is preserved during a cliff event so the detection re-triggers on subsequent cycles until heartbeats recover.
- Uses a dedicated HEARTBEAT_CLIFF_THRESHOLD config key, independent of the existing THRESHOLD used for aggregate failure gating.

Also includes:
- Prometheus metrics: k8s_api_reachable gauge, heartbeat_cliff_total counter, heartbeat_rejected_total counter
- Prometheus monitoring guide with alert rules, Grafana dashboard queries, PodMonitor/ScrapeConfig setup, and telemetry-operator integration
- Documentation updates for architecture, deployment guide, and configuration reference
- Tests for K8s partition detection, heartbeat cliff detection, and credential namespace fallback
- Fix operator precedence in controller finalizer guard
- Fix flaky concurrent cache refresh test (race in mock setup)


templates/instanceha/config/config.yaml
dciabrin, line 1: Might be nice to add comment into each key in a subsequent PR, now that we have a lot of options, this would likely help clarify impact of those ones.

dciabrin (approved):
LGTM, seems fine and sufficiently important so that we should merge that sooner than later.

PR #596, opened 2026-06-08
Final InstanceHA cleanup and small fixes

lmiccini:
This PR consists of three commits:

**Commit 1: Lazy-import novaclient and keystoneauth1 SDK modules**

Move novaclient and keystoneauth1 imports from module level into the functions that use them (nova_login, _server_evacuate, _handle_nova_exception, main). 
Eliminates ~135 lines of mock exception boilerplate across 4 test files. 
Add shared patch_pipeline() fixture in conftest.py replacing 7-deep nested patch blocks in process_service tests. 
Bind conftest sys.modules submodule mocks to parent mock attributes so lazy imports and patch() targets resolve to the same objects.

**Commit 2: Remove dead code and consolidate redundant patterns**

Remove MigrationStatus enum (inline strings), dead _previous_hash state, unreachable VALIDATION_PATTERNS entries, unused _check_critical_services params, redundant validate_input in _host_fence, single-use _cleanup_dict_by_condition helper, and redundant requests.exceptions import. 
Consolidate 5 manual error+traceback log pairs into _safe_log_exception calls. 
Deduplicate _get_nova_connection by delegating to _establish_nova_connection.

**Commit 3: Fix fail-open startup, fencing decision log, and document safety model**

Initialize k8s_api_reachable=True (fail-open) to prevent startup deadlock where fencing was silently blocked until the background health check thread completed its first probe. 
Initialize K8S_API_REACHABLE Prometheus gauge to 1 to match. 
Emit fencing decision log line before early return so cliff-detection and all-heartbeat-alive cases are logged. 
Warn at startup when CHECK_HEARTBEAT is disabled. 
Add Fencing Safety Model section to architecture doc covering gate chain, fail-open/fail-closed semantics, startup race window trade-off, and MedIK8s/InstanceHA responsibility boundary.

### DIFF
--- a/docs/instanceha_architecture.md
+++ b/docs/instanceha_architecture.md
@@ -9,21 +9,22 @@ InstanceHA is a high-availability service for OpenStack that automatically detec
 1. [Core Components](#core-components)
 2. [Architecture Patterns](#architecture-patterns)
 3. [Service Workflow](#service-workflow)
-4. [Critical Services Validation](#critical-services-validation)
-5. [Configuration System](#configuration-system)
-6. [Evacuation Mechanisms](#evacuation-mechanisms)
-7. [Fencing Agents](#fencing-agents)
-8. [Kubernetes Events and Prometheus Metrics](#kubernetes-events-and-prometheus-metrics)
-9. [Advanced Features](#advanced-features)
-10. [Region Handling](#region-handling)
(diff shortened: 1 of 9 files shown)


stuggi (approved):
/lgtm

PR #597, opened 2026-06-08
Propagate RabbitMQ LoadBalancerClass override

SeanMooney:
RabbitMq service overrides accept spec.loadBalancerClass, but the
client-facing Service renderer did not propagate it to the generated
Service.

On MicroShift, unclassed LoadBalancer Services can be claimed by the
built-in LoadBalancer controller, causing RabbitMQ DNS data to expose the
node IP instead of the MetalLB VIP expected by dataplane nodes.

This PR keeps the two-commit green-green flow:
- document the existing behavior with a FIXME unit test
- propagate loadBalancerClass and flip the assertion

Validation:
- go test ./internal/rabbitmq -count=1

Closes: [OSPRH-30887](https://redhat.atlassian.net/browse/OSPRH-30887)

### DIFF
--- a/internal/rabbitmq/service.go
+++ b/internal/rabbitmq/service.go
@@ -26,6 +26,14 @@ func serviceOverrideIPFamilyPolicy(r *rabbitmqv1.RabbitMq) *corev1.IPFamilyPolic
 	return nil
 }
 
+// serviceOverrideLoadBalancerClass returns the LoadBalancerClass from override, or nil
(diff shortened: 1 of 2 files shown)


lmiccini:
let's close this since we are mergin https://github.com/openstack-k8s-operators/infra-operator/pull/598 ?

SeanMooney:
yep #598 is better thanks for proposing it

PR #604, opened 2026-06-15
Skip NodeSet sync for controlplane-only credential rotation

lmiccini:
During RabbitMQ credential rotation, the TransportURL controller uses a two-phase NodeSet hash-sync check to ensure EDPM nodes pick up new credentials before releasing the old user. This is only needed for services that run agents on EDPM nodes (Nova, Neutron).

Determine EDPM status from the TransportURL's ownerReference Kind: if the owner is Nova or NeutronAPI, wait for NodeSet deployment; otherwise release the old user immediately. The edpm-service annotation is available as an explicit override for edge cases.


stuggi (approved):
/lgtm

PR #606, opened 2026-06-16
Immutable secrets and consumer finalizers for TransportURL credential rotation

lmiccini:
During credential rotation, create a new immutable secret instead of patching the existing mutable one. Consuming service operators add a consumer finalizer to the new secret and remove it from the old one after completing their rollout. The TransportURL controller waits for all consumer finalizers to be removed before deleting the old secret and releasing the old RabbitMQ user.

This follows the Keystone Application Credentials pattern already established across the ecosystem and eliminates the need for cross-operator unstructured lookups (17-API-group RBAC removed).

Backward compatible: if no consumer finalizer is present on the old secret (pre-upgrade operator), the old user is released immediately.

EDPM path unchanged: continues using NodeSet hash-sync gating.


lmiccini:
recheck

PR #608, opened 2026-06-22
Immutable secrets and consumer finalizers for TransportURL credential rotation

lmiccini:
Enable safe credential rotation for TransportURL by creating immutable, content-hashed secrets and coordinating cleanup across consuming operators.
    
    Key changes:
    - Create immutable transport secrets (rabbitmq-transport-url-{name}-{hash})
      during rotation so consumers cannot accidentally mutate credentials mid-rollout
    - Add per-consumer finalizers (turl.openstack.org/t-{name}) on shared
      RabbitMQUser and RabbitMQVhost CRs to track which operators are still
      using the old credentials
    - Let consuming operators (e.g. Nova, Heat) signal rollout completion by
      removing their finalizer, so TransportURL knows when it is safe to
      delete the old user
    - Gate old user cleanup on both consumer finalizer removal and NodeSet
      secret hash synchronization — if the secret is not tracked by the
      dataplane, release immediately; otherwise wait for a full NodeSet deploy
    - Prevent SecretName flip-flop by comparing content hashes before creating
      a new immutable secret
    - Auto-delete orphaned RabbitMQUser CRs once all consumers have released
      their finalizers


lmiccini:
recheck

lmiccini:
recheck

lmiccini:
recheck

apis/rabbitmq/v1beta1/transporturl_types.go
stuggi: The NodeSetSynced field is still present in the status type and is still cleared in releasePendingUser, but the two-phase logic that set it to "false" has been removed. It's only ever set to "" now. Is it still needed?

internal/controller/rabbitmq/transporturl_controller.go
```diff
@@ -628,6 +700,85 @@ func (r *TransportURLReconciler) createTransportURLSecret(
...
+		Data:      template.Data,
+		Immutable: ptr.To(true),
+	}
+	if err := controllerutil.SetControllerReference(instance, secret, r.Scheme); err != nil {
+		return nil, "", fmt.Errorf("failed to set controller reference on transport secret: %w", err)
+	}
+
+	if err := r.Create(ctx, secret); err != nil {
+		if !k8s_errors.IsAlreadyExists(err) {
+			return nil, "", fmt.Errorf("failed to create immutable transport secret %s: %w", secret.Name, err)
```
stuggi, line 743: When r.Create(ctx, secret) returns AlreadyExists, the function returns the local secret object (no server metadata) and the locally-computed contentHash. If a prior reconcile already created the secret but the status patch failed, this works by accident because the name is deterministic. However, it would be cleaner to Get the existing secret on AlreadyExists to confirm the data matches, avoiding a silent mismatch if hash collisions ever occur:

```
  if err := r.Create(ctx, secret); err != nil {
      if !k8s_errors.IsAlreadyExists(err) {
          return nil, "", fmt.Errorf("failed to create immutable transport secret %s: %w", secret.Name, err)
      }
      // Fetch the existing secret to confirm content matches
      if err := r.Get(ctx, types.NamespacedName{Name: secret.Name, Namespace: secret.Namespace}, secret); err != nil {
          return nil, "", fmt.Errorf("failed to get existing immutable transport secret %s: %w", secret.Name, err)
      }
  }
```

internal/controller/rabbitmq/transporturl_controller.go
```diff
@@ -518,29 +518,77 @@ func (r *TransportURLReconciler) reconcileNormal(ctx context.Context, instance *
...
+				condition.SeverityInfo,
+				rabbitmqv1.TransportURLReadyInitMessage))
+			return ctrl.Result{RequeueAfter: time.Second * 5}, nil
+		}
+		if !skipSecretNameFlip {
+			if instance.Status.SecretName != "" && instance.Status.SecretName != secret.Name {
+				instance.Status.PreviousSecretName = instance.Status.SecretName
+			}
+			instance.Status.SecretName = secret.Name
+			instance.Status.SecretHash = hash
```
stuggi, line 588: This sets PreviousSecretName without a corresponding PreviousRabbitmqUserRef, which then triggers the orphan cleanup block at line 624. This block checks for consumer finalizers on a secret that was never part of a consumer-driven rotation. The state machine has two independent cleanup paths (lines 608-618 for rotation, lines 624-642 for orphans) that overlap and can interact in unexpected ways. Should we either:
  - Not setting PreviousSecretName in the normal path (only set it during rotation), or
  - Adding a comment explaining this is intentional for cleaning up stale immutable secrets post-rotation?

stuggi (approved):
/lgtm

PR #615, opened 2026-06-27
Fix proxy sigterm handling in amqp-proxy, enable CMR and override terminationGracePeriodSeconds

lmiccini:
This PR addresses three issues/shortcomings that emerged during the rabbitmq upgrade testing:

  - Handle SIGTERM in AMQP proxy: The Python proxy sidecar only handled SIGINT, not SIGTERM. Since kubelet sends
  SIGTERM to stop containers, the proxy hung indefinitely preventing pod termination.

-> this is needed so that rolling the pods after annotating the cluster to remove the proxy works without having to force-delete the pods.

  - Enable Continuous Membership Reconciliation (CMR): Quorum queues can end up under-replicated when clients
  declare queues while not all nodes are available (rolling restarts, parallel pod startup after data-wipe
  upgrades). RabbitMQ does not retroactively grow membership. CMR (RabbitMQ 4.x) automatically grows quorum queues to the target group size — the trigger fires within 10s of a node joining, with a 60-minute periodic safety
  net.

-> this minimizes the chances that the prestophook would wait indefinitely for all the queues to be replicated across the cluster before draining the node during stop or rolling restarts.

  - Migrate terminationGracePeriodSeconds from 604800 to 60: The old 7-day default inherited from the upstream
  cluster-operator causes pods to stay in Terminating state for days when combined with containers that don't
  handle SIGTERM. The defaulting webhook migrates existing clusters still carrying the old value; custom values
  are preserved.

self explanatory, we want to align existing clusters to the new default to avoid getting stuck indefinitely if we hit any of the situations above ^

### DIFF
--- a/apis/rabbitmq/v1beta1/rabbitmq_webhook.go
+++ b/apis/rabbitmq/v1beta1/rabbitmq_webhook.go
@@ -140,6 +140,16 @@ func (spec *RabbitMqSpecCore) Default(isNew bool) {
 		spec.QueueType = &queueType
 	}
 
+	// Migrate terminationGracePeriodSeconds from old default (604800) to new (60).
+	// The old value was inherited from rabbitmq-cluster-operator and causes pods
+	// to get stuck in Terminating state for up to a week.
+	oldGracePeriod := int64(604800)
+	newGracePeriod := int64(60)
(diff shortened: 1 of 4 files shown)


stuggi (approved):
/lgtm

PR #621, opened 2026-07-10
Decouple detection from evacuation and evaluate safety gates cumulatively across poll cycles

lmiccini:
The main poll loop previously blocked during evacuation, preventing detection of new host failures until in-flight evacuations completed. This change decouples detection (admission) from processing (evacuation) and makes every cumulative safety gate evaluate total cluster impact.

Async processing:
- Add persistent ThreadPoolExecutor to InstanceHAService for background host processing (fence + evacuate)
- Rename _process_stale_services to _admit_stale_services: safety gates run synchronously, then accepted hosts are submitted to the executor non-blocking
- Main loop no longer blocks on evacuation -- polls every POLL seconds regardless of in-flight work

Total-impact safety gates:
- Global threshold: evaluates total cluster health (all down/forced_down hosts vs all services) using _count_down_hosts. Catches slow cascades where hosts fail one per cycle -- previously-evacuated hosts that haven't recovered count as impacted
- All-stale check: includes in-flight count when checking whether all active services appear stale
- Per-aggregate threshold: counts in-flight hosts toward their aggregate failure budget
- Heartbeat cliff: excludes hosts_processing from active heartbeat counts so expected heartbeat absence does not trigger false cliff detection

Processing health:
- Liveness probe detects stale processing entries (oldest entry > 2x max(FENCING_TIMEOUT, EVACUATION_TIMEOUT)) since the main loop no longer blocks and cannot detect hung processing via poll staleness
- Graceful shutdown: SIGTERM stops new submissions, then waits for in-flight evacuations to complete

### DIFF
--- a/docs/instanceha_architecture.md
+++ b/docs/instanceha_architecture.md
@@ -205,6 +205,12 @@ def __init__(self, config_manager: ConfigManager, cloud_client: Optional[OpenSta
     # Host processing tracking
     self.hosts_processing = defaultdict(float)
     self.processing_lock = threading.Lock()
+    try:
+        workers = self.config.get_config_value('WORKERS')
+    except (TypeError, ValueError, AttributeError):
+        workers = 4
+    self.processing_executor = concurrent.futures.ThreadPoolExecutor(
+        max_workers=workers if isinstance(workers, int) and workers > 0 else 4)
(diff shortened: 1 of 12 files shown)


antonioromito (approved):
good to go!

PR #634, opened 2026-07-24
Resolve InstanceHa container image from ConfigMap before fallback

lmiccini:
The webhook was always defaulting spec.containerImage, which meant the controller never reached the ConfigMap lookup path. This made the infra-instanceha-config ConfigMap (managed by openstack-operator via OpenStackVersion) effectively dead code.

Remove the ContainerImage webhook defaulting and reorder the resolution in GetContainerImage() so the priority is:
1. spec.containerImage (explicit user override)
2. infra-instanceha-config ConfigMap (openstack-operator managed)
3. RELATED_IMAGE env var / hardcoded fallback (standalone deployment)

Jira: https://redhat.atlassian.net/browse/OSPRH-33358

### DIFF
--- a/apis/instanceha/v1beta1/instanceha_webhook.go
+++ b/apis/instanceha/v1beta1/instanceha_webhook.go
@@ -58,9 +58,6 @@ func (r *InstanceHa) Default() {
 
 // Default - set defaults for this InstanceHa spec
 func (spec *InstanceHaSpec) Default() {
-	if spec.ContainerImage == "" {
-		spec.ContainerImage = instanceHaDefaults.ContainerImageURL
-	}
 	if spec.OpenStackCloud == "" {
 		spec.OpenStackCloud = OpenStackCloud
 	}
--- a/internal/controller/instanceha/instanceha_controller.go
+++ b/internal/controller/instanceha/instanceha_controller.go
@@ -1005,33 +1005,30 @@ func (r *Reconciler) findInstanceHaWithPendingRotation(ctx context.Context, _ cl
 	return requests
 }
 
-// GetContainerImage returns the container image to use for the instance, either from
-// the provided containerImage parameter or from the infra-instanceha-config ConfigMap
+// GetContainerImage returns the container image for InstanceHa. Priority:
+// 1. spec.containerImage (explicit user override)
+// 2. infra-instanceha-config ConfigMap (managed by openstack-operator)
+// 3. RELATED_IMAGE env var or hardcoded fallback (standalone deployment)
 func (r *Reconciler) GetContainerImage(
 	ctx context.Context,
 	containerImage string,
 	src client.Object,
 ) (string, error) {
-	cm := &corev1.ConfigMap{}
-	instanceHaConfigMapName := "infra-instanceha-config"
-
 	if len(containerImage) > 0 {
 		return containerImage, nil
 	}
 
-	objectKey := client.ObjectKey{Namespace: src.GetNamespace(), Name: instanceHaConfigMapName}
-	err := r.Get(ctx, objectKey, cm)
-	if err != nil {
-		return "", err
+	cm := &corev1.ConfigMap{}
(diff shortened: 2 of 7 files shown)


lmiccini:
recheck

internal/controller/instanceha/instanceha_controller.go
```diff
@@ -1005,33 +1005,26 @@ func (r *Reconciler) findInstanceHaWithPendingRotation(ctx context.Context, _ cl
...
-
-	if cm.Data == nil {
-		return util.GetEnvVar("RELATED_IMAGE_INFRA_INSTANCE_HA_IMAGE_URL_DEFAULT", ""), nil
-	}
-
-	if cmImage, exists := cm.Data["instanceha-image"]; exists {
-		return cmImage, nil
+	cm := &corev1.ConfigMap{}
+	objectKey := client.ObjectKey{Namespace: src.GetNamespace(), Name: "infra-instanceha-config"}
+	if err := r.Get(ctx, objectKey, cm); err == nil && cm.Data != nil {
```
stuggi: ok do ignore err != nil?

internal/controller/instanceha/instanceha_controller.go
```diff
@@ -1005,33 +1005,26 @@ func (r *Reconciler) findInstanceHaWithPendingRotation(ctx context.Context, _ cl
...
+		if cmImage, exists := cm.Data["instanceha-image"]; exists && len(cmImage) > 0 {
+			return cmImage, nil
+		}
 	}
 
-	return "", fmt.Errorf("no container image found: key 'instanceha-image' not in ConfigMap %s and RELATED_IMAGE_INFRA_INSTANCE_HA_IMAGE_URL_DEFAULT not set", instanceHaConfigMapName)
+	return util.GetEnvVar("RELATED_IMAGE_INFRA_INSTANCE_HA_IMAGE_URL_DEFAULT", instancehav1.InstanceHaContainerImage), nil
```
stuggi: https://github.com/openstack-k8s-operators/infra-operator/blob/main/apis/instanceha/v1beta1/instanceha_types.go#L234 and  https://github.com/openstack-k8s-operators/infra-operator/blob/main/config/default/manager_default_images.yaml#L21 use `RELATED_IMAGE_INSTANCE_HA_IMAGE_URL_DEFAULT`. is it expected that we have `RELATED_IMAGE_INFRA_...` and `RELATED_IMAGE_INSTANCE_HA_...`?

apis/instanceha/v1beta1/instanceha_webhook.go
```diff
@@ -58,9 +58,6 @@ func (r *InstanceHa) Default() {
 
 // Default - set defaults for this InstanceHa spec
 func (spec *InstanceHaSpec) Default() {
-	if spec.ContainerImage == "" {
-		spec.ContainerImage = instanceHaDefaults.ContainerImageURL
```
stuggi, line 62: the comment in 
https://github.com/openstack-k8s-operators/infra-operator/blob/main/apis/instanceha/v1beta1/instanceha_types.go#L80 seems now no longer true `will be set to environmental default if empty`

internal/controller/instanceha/instanceha_controller.go
lmiccini: fixed (removed _INFRA_)

apis/instanceha/v1beta1/instanceha_webhook.go
lmiccini: amended comment

internal/controller/instanceha/instanceha_controller.go
lmiccini: the idea was to ignore the error and gracefully move on to the fallback. w/e anyone will ever install infra without openstack-operator is ofc debatable.

stuggi (approved):
/lgtm

lmiccini:
recheck

PR #638, opened 2026-07-27
Resume-safe evacuation: check migration status before re-evacuating on pod restart

lmiccini:
When the InstanceHA pod is killed mid-evacuation (e.g. by a liveness probe failure), on restart the resume path blindly re-evacuated all servers, disrupting in-progress migrations by resetting task_state. This change makes the resume path migration-aware: it checks each server's migration status before acting, monitors in-progress migrations, skips completed ones, and only re-evacuates when the migration errored or doesn't exist. Also includes REBUILD status servers in the evacuable list on resume so mid-evacuation VMs are not silently dropped.


stuggi (approved):
/lgtm

PR #643, opened 2026-07-28
Prevent unnecessary RabbitMQ rolling restarts during scaling and cert rotation

lmiccini:
Three changes to align with the upstream rabbitmq-cluster-operator and prevent quorum queue corruption caused by unnecessary rolling restarts:

1. Remove TLS secret data and replica count from config hash

   Stop hashing TLS secret content (tlsSecretData, caSecretData) and the replica count into the pod-template annotation. The secret names are still hashed so that switching to a different cert triggers a restart, but content changes (cert-manager re-issuance, CA chain updates, rotation) no longer cause rolling restarts. TLS certificates are mounted via projected volumes that kubelet updates in-place, and RabbitMQ 4.x reloads them from disk automatically.

   During initial deployment the playbook scales RabbitMQ from 1 to 3 replicas, which adds new DNS SANs to the Certificate CR and triggers cert-manager to re-issue. When secret data was hashed, this caused a rolling restart before quorum queue replicas had synced, leading to Raft log divergence and a permanent crash loop (ra_server assertion "PLIdx < LastApplied").

2. Make PreStop hook replica-count-independent

   The PreStop command previously varied based on spec.replicas (skipping quorum checks for single-replica clusters). This meant scaling from 1 to 3 changed the pod template, triggering a restart of existing pods. Now the quorum check is conditional at runtime: it queries rabbit_nodes:list_running() and only runs await_online_quorum_plus_one when more than one node is active.

   Timeouts are reduced from 600s to 10s for quorum checks and 30s for drain, fitting within the 60s terminationGracePeriodSeconds. If quorum is already satisfied (normal rolling restart), the check passes instantly; in degraded scenarios it times out quickly rather than blocking.

3. Set explicit UpdateStrategy on the StatefulSet

   Always set RollingUpdate with Partition=0, matching the upstream cluster-operator. This prevents stale partition values (from migration or manual debugging) from silently blocking pod updates.

### DIFF
--- a/internal/controller/rabbitmq/rabbitmq_controller.go
+++ b/internal/controller/rabbitmq/rabbitmq_controller.go
@@ -388,45 +388,22 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ct
 	}
 
 	// Calculate hash for config tracking.
-	// Include TLS secret content so that certificate rotation (e.g. by
-	// cert-manager) triggers a rolling restart of the StatefulSet pods.
+	// TLS cert content is NOT included: certs are mounted via projected
+	// volumes that kubelet updates in-place, and RabbitMQ 4.x reloads
+	// them automatically. Hashing cert data here would trigger a
+	// rolling restart on every cert-manager update, which can corrupt
+	// quorum queues when replicas haven't fully synced yet.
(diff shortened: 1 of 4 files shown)


lmiccini:
recheck

lmiccini:
recheck

PR #649, opened 2026-07-29
Add Explicit Private Key Stanzas for Metrics and Webhook Services Configured with Cert Manager

subecho:
Previously, we did not specify the parameters of the private keys that would generated for the Metrics and Webhook services in the Infra Operator, relying instead on whatever the defaults might be.

As we move towards enabling post-quantum safety in our operators, we are starting by explicitly stating what the private key algorithm and key size shall be. This is to ensure that the keys are exactly what we want them to be and to provide an easy-to-configure place to change the algorithm and key size should the need arise (such as when quantum safe algorithms become available to use for generating private keys.)

### DIFF
--- a/config/certmanager/certificate-metrics.yaml
+++ b/config/certmanager/certificate-metrics.yaml
@@ -21,4 +21,3 @@ spec:
   privateKey:
     algorithm: ECDSA
     size: 384
-
--- a/config/certmanager/certificate-metrics.yaml
+++ b/config/certmanager/certificate-metrics.yaml
@@ -18,3 +18,7 @@ spec:
     kind: Issuer
     name: selfsigned-issuer
   secretName: metrics-server-cert
+  privateKey:
+    algorithm: ECDSA
+    size: 384
+
--- a/config/certmanager/certificate-webhook.yaml
+++ b/config/certmanager/certificate-webhook.yaml
@@ -21,4 +21,3 @@ spec:
   privateKey:
     algorithm: ECDSA
     size: 384
-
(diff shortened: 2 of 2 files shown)


stuggi:
we could make this change in the infra-operator, but to change this in the product we need https://github.com/openstack-k8s-operators/openstack-operator/pull/1973 . best we sync with @fmount when he is back from pto on which size we use.

subecho:
> we could make this change in the infra-operator, but to change this in the product we need [openstack-k8s-operators/openstack-operator#1973](https://github.com/openstack-k8s-operators/openstack-operator/pull/1973) . best we sync with @fmount when he is back from pto on which size we use.

Makes sense to me! I was using the recommendations from the PQC scans that we did a while ago which looks to be the same ones being utilized in [openstack-k8s-operators/openstack-operator#1973](https://github.com/openstack-k8s-operators/openstack-operator/pull/1973). Happy to help out however I can! Thanks @stuggi!

fmount:
> > we could make this change in the infra-operator, but to change this in the product we need [openstack-k8s-operators/openstack-operator#1973](https://github.com/openstack-k8s-operators/openstack-operator/pull/1973) . best we sync with @fmount when he is back from pto on which size we use.
> 
> Makes sense to me! I was using the recommendations from the PQC scans that we did a while ago which looks to be the same ones being utilized in [openstack-k8s-operators/openstack-operator#1973](https://github.com/openstack-k8s-operators/openstack-operator/pull/1973). Happy to help out however I can! Thanks @stuggi!

Hey @subecho thanks for this. In theory this change makes sense because aligns the OLM deployment part of infra-operator w/ the final configuration we want to achieve.
However, as @stuggi pointed out, except for openstack-operator we deploy all the operators from it by rendering the required assets. This makes this patch less important but it still makes sense.
Let's align first on the long term goal so we can maybe plan to do the same across the openstack-operators.

PR #651, opened 2026-07-31
Simplify RabbitMQ PDB handling using lib-common helpers

lmiccini:
Use pdb.DeletePDBWithName() instead of manual PDB object construction and deletion. Also handle ctrlResult from CreateOrPatch to properly support requeueing.


stuggi (approved):
/lgtm

stuggi:
recheck

PR #657, opened 2026-08-11
Use absolute memory watermark from container limit for RabbitMQ

lmiccini:
RabbitMQ nodes can't always detect cgroup memory limits in containers, so an absolute watermark is recommended over a relative one (see rabbitmq.com/docs/memory). Compute 60% of the pod's memory limit and set vm_memory_high_watermark.absolute instead of the previous hardcoded vm_memory_high_watermark.relative = 0.6. Falls back to relative if no memory limit is set.

Jira: https://redhat.atlassian.net/browse/OSPRH-34633


abays (approved):
/lgtm

PR #669, opened 2026-08-20
Source workload UIDs/GIDs from lib-common users registry

lmiccini:
Replace the per-package UID/GID constants for rabbitmq, redis and memcached with the central registry in lib-common's users module, which is the single source of truth mirroring TCIB's uid_gid_manage.

This also fixes rabbitmq's pinned UID: it was hardcoded to 999 (the kolla/docker-official value), whereas the s2i (tcib) image assigns the rabbitmq user 42439. The main container, pod-level FSGroup/RunAsUser and the amqp-proxy sidecar now use users.RabbitmqUID/RabbitmqGID. redis and memcached already matched the registry values (42460 / 42457) and are switched over to keep them from drifting from the image builds.


lmiccini:
recheck

stuggi (approved):
/lgtm

PR #671, opened 2026-08-26
Add Redis/Valkey compatibility to redis templates

lmiccini:
This change makes the redis operator templates compatible with both Redis and Valkey container images without requiring operator code changes.

Changes:
- Remove tar dependency from config generation, replacing it with find + cp for better portability
- Auto-detect redis-cli/redis-server or valkey-cli/valkey-server binaries at runtime and use the appropriate commands
- Update startup scripts to use detected binary variables instead of hardcoded redis-server/redis-sentinel commands
- Change data directory from /var/lib/redis to /var/lib/config-data/generated to use the writable EmptyDir volume (valkey user lacks permissions for /var/lib/redis)

The operator continues to work with existing Redis images while also supporting Valkey as a drop-in replacement.

### DIFF
--- a/internal/redis/volumes.go
+++ b/internal/redis/volumes.go
@@ -46,6 +46,12 @@ func getVolumes(r *redisv1.Redis) []corev1.Volume {
 			},
 		},
 		{
+			Name: "redis-data",
+			VolumeSource: corev1.VolumeSource{
+				EmptyDir: &corev1.EmptyDirVolumeSource{},
+			},
+		},
+		{
 			Name: "config-data",
 			VolumeSource: corev1.VolumeSource{
 				ConfigMap: &corev1.ConfigMapVolumeSource{
@@ -140,6 +146,9 @@ func getRedisVolumeMounts(r *redisv1.Redis) []corev1.VolumeMount {
 		MountPath: "/var/lib/operator-scripts",
 		ReadOnly:  true,
 		Name:      "operator-scripts",
+	}, {
+		MountPath: "/var/lib/redis",
+		Name:      "redis-data",
 	}}
 	vm = append(vm, getTLSVolumeMounts(r)...)
 	return vm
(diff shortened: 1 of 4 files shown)


dciabrin:
If I read that correctly, we may have the same rolling update problem that we had in mariadb-operator.

So when the redis-operator gets updated, it recreates config-maps with templates that now points to /var/lib/config-data/generated.
redis-2 pod will get recreated, so it picks up the new scripts and template, and generates the right thing in the new directory.
However if for any reason redis-0 and redis-1 get restart before redis-2 is restarted, redis-2 won't be able to go back to ready state until redis-0/1 restart and get ready, so the rolling update with will on redis-0/1 to finish.
And I'm not if redis-0/1 pick up the new templates/scripts versions or if they restart with the previous version (because the pods are not recreated). So we may have a deadlock during the rolling update.

lmiccini:
recheck

lmiccini:
> If I read that correctly, we may have the same rolling update problem that we had in mariadb-operator.
> 
> So when the redis-operator gets updated, it recreates config-maps with templates that now points to /var/lib/config-data/generated. redis-2 pod will get recreated, so it picks up the new scripts and template, and generates the right thing in the new directory. However if for any reason redis-0 and redis-1 get restart before redis-2 is restarted, redis-2 won't be able to go back to ready state until redis-0/1 restart and get ready, so the rolling update with will on redis-0/1 to finish. And I'm not if redis-0/1 pick up the new templates/scripts versions or if they restart with the previous version (because the pods are not recreated). So we may have a deadlock during the rolling update.

this version of the PR should be ok, since we did not change any path that would cause rolling update issues. we just added a new volume mount and changed the config generation method, but the end result should be the same. even if the pods would restart they should still be able to pick things up from where they left?

antonioromito (approved):

PR #678, opened 2026-09-10
Fix RabbitMq sample to use spec.persistence.storageClassName

lmiccini:
## Problem

The RabbitMq spec no longer exposes a top-level `storageClass` field — the storage class is now configured under **`spec.persistence.storageClassName`** (`apis/rabbitmq/v1beta1`, `DeprecatedPersistenceSpec.StorageClassName`). But `config/samples/rabbitmq_v1beta1_rabbitmq.yaml` still sets the removed top-level `spec.storageClass`.

Because `spec.storageClass` is now an unknown field, the API server **prunes** it, so a CR based on this sample gets **no** storage class and its PVC falls back to the cluster **default** StorageClass instead of `local-storage`.

## Impact

On OpenShift-on-OpenStack CI clusters the default StorageClass is `standard-csi` (Cinder), while sibling components (galera, OVN) use `local-storage`. This made RabbitMQ the only Cinder-CSI workload, exposing it to `ControllerPublishVolume` attach timeouts that wedged `rabbitmq-server-0` in `Init:0/1` and cascaded into consumer operators' `build-deploy-kuttl` failures (they hang `Waiting for TransportURL ... secret to be created`).

## Fix

Update the sample to set the storage class under `spec.persistence.storageClassName`, matching the current API.

Companion fix for install_yamls (which templates this CR for kuttl): openstack-k8s-operators/install_yamls#1183.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


stuggi (approved):
/lgtm

